PR #2, opened 2018-10-22
Implement xpython_logger class

martinRenou:
This class allows to redirect messages received from print() in python
to logger functions of type std::function<void(std::string&)>

### DIFF
--- a/src/xpython_interpreter.cpp
+++ b/src/xpython_interpreter.cpp
@@ -148,16 +148,23 @@ namespace xpyt
         py::module xeus_python_logger = py::module::import("xeus_python_logger");
 
         // Create XPythonLogger instance
-        py::object logger = xeus_python_logger.attr("XPythonLogger")();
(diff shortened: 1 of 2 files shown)


src/xpython_interpreter.cpp
```diff
@@ -159,7 +136,35 @@ namespace xpyt
...
+            this->publish_stream("stderr", message);
+        };
+
+        // Add publish_stream as a logger function
+        out_logger.attr("add_logger")(publish_stdout_stream);
+        err_logger.attr("add_logger")(publish_stderr_stream);
+
+        // And replace sys.stdout by the XPythonLogger instance
+        sys.attr("stdout") = out_logger;
+        sys.attr("stderr") = err_logger;
```
Carreau: You may want to save the original ones in `__std*__`

martinRenou: Good point, thanks!

PR #3, opened 2018-10-23
Implement xpython_logger class

martinRenou:
This class allows to redirect messages received from print() in python
to logger functions of type std::function<void(std::string&)>

### DIFF
--- a/src/xpython_interpreter.cpp
+++ b/src/xpython_interpreter.cpp
@@ -35,10 +35,7 @@ namespace xpyt
         redirect_output();
     }
 
-    xpython_interpreter::~xpython_interpreter()
-    {
-        py::finalize_interpreter();
-    }
+    xpython_interpreter::~xpython_interpreter() {}
 
     xeus::xjson xpython_interpreter::execute_request_impl(
         int execution_counter,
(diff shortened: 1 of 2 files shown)


src/xpython_interpreter.cpp
```diff
@@ -159,7 +136,39 @@ namespace xpyt
...
+            this->publish_stream("stderr", message);
+        };
+
+        // Add publish_stream as a logger function
+        out_logger.attr("add_logger")(publish_stdout_stream);
+        err_logger.attr("add_logger")(publish_stderr_stream);
+
+        // Save current stdout/stderr
+        sys.attr("__stdout__") = sys.attr("stdout");
+        sys.attr("__stderr__") = sys.attr("stderr");
```
martinRenou: @Carreau are you sure this is needed? From the [documentation](https://docs.python.org/2/library/sys.html#sys.__stderr__):

> These objects contain the original values of stdin, stderr and stdout at the start of the program. They are used during finalization, and could be useful to print to the actual standard stream no matter if the sys.std* object has been redirected.

It seems to me that `__stdout__` and `__stderr__` already contains the `stdout` and `stderr`

PR #11, opened 2018-11-06
Enable CI

martinRenou:
Enable CI and fix dependencies, do some cleaning

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,65 +36,12 @@ configure_file (
     "${CMAKE_CURRENT_SOURCE_DIR}/share/jupyter/kernels/xeus-python/kernel.json"
 )
 
-#######################
-# Rely on llvm-config #
-#######################
-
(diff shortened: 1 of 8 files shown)


CMakeLists.txt
```diff
@@ -146,7 +146,7 @@ set(XEUSCLING_SRC
     src/xpyt_config.hpp
 )
 
-set(LIBS ${PYTHON_LIBRARIES} clingMetaProcessor clingUtils xeus pugixml cxxopts::cxxopts)
+set(LIBS ${PYTHON_LIBRARIES} xeus pugixml)
```
SylvainCorlay: I don't think you need pugixml either for xeus-python.

SylvainCorlay:
cf https://github.com/nlohmann/json/issues/1337

`Python.h` defines `snprintf` as a macro. Including `xjson` before `pybind` or `Python` should fix it.

PR #17, opened 2018-11-13
WIP: Implement display repr

martinRenou:


SylvainCorlay:
From the documentation of IPython, it is clear that `_repr_mimebundle_` takes precedence over all others.

However, I am unsure about the order of precedence between `_repr_html_`, `_repr_json_`, `_repr_jpeg_`, `_repr_png_`, `_repr_svg_`, `_repr_latex_`.

src/xinterpreter.cpp
```diff
@@ -240,15 +240,60 @@ namespace xpyt
...
+                }
+                else if (hasattr(obj, "_repr_latex_"))
+                {
+                }
+                else if (hasattr(obj, "__repr__"))
+                {
+                    pub_data["text/plain"] = static_cast<std::string>(
+                        py::str(obj.attr("__repr__")())
+                    );
+                }
```
SylvainCorlay: Should we check if it has `ipython_display` hook and say that it is not supported with a warning?

src/xinterpreter.cpp
```diff
@@ -240,15 +240,60 @@ namespace xpyt
...
+
+                std::cout << pub_data << std::endl;
```
SylvainCorlay: leftover

src/xinterpreter.cpp
```diff
@@ -240,15 +240,60 @@ namespace xpyt
...
         m_displayhook = xeus_python_display.attr("XPythonDisplay")();
 
-        py::cpp_function publish_display = [this](int execution_counter, py::object obj){
+        py::cpp_function publish_display = [this, py_json](int execution_counter, py::object obj){
             if (!obj.is_none())
             {
                 xeus::xjson pub_data;
-                pub_data["text/plain"] = py::str(obj);
+
+                if (hasattr(obj, "_repr_mimebundle_"))
```
SylvainCorlay: TODO: separate in get_display_pub_data (or other name) function, to be called both from the display hook and the explicit `display` and `display_id` functions.

src/xinterpreter.cpp
```diff
@@ -253,7 +253,9 @@ namespace xpyt
         };
 
         m_displayhook.attr("add_hook")(publish_display);
-
         sys.attr("displayhook") = m_displayhook;
+
+        // Create a global "display" function
+        py::globals()["display"] = m_displayhook;
```
martinRenou: I guess it will be more complicated than that. Because the `display` function from `IPython` takes some arguments that could be useful in some cases?

Carreau:
> From the documentation of IPython, it is clear that _repr_mimebundle_ takes precedence over all others.

I think the display order and protocol in IPython is a mess,  if you think that some of our decision are wrong feel free to attempt other things.  In particular I would be happy to have something more coherent, like always having the `_repr_` return the `(data,metadata)` tuple (I know it's hard to enforce).  Also if you could solve [this](https://github.com/ipython/ipython/pull/10651) instead of triggering side-effect directly that would be nice. 

> However, I am unsure about the order of precedence between `_repr_html_`, `_repr_json_`, `_repr_jpeg_`, `_repr_png_`, `_repr_svg_`, `_repr_latex_`.

None that you care about. We compute all and send them all. The frontends choose. They are also all embedded in the notebook for nbconvert.

src/xdisplay.cpp
```diff
@@ -42,6 +46,63 @@ namespace xpyt
...
+                py::str(obj.attr("_repr_html_")())
+            );
+        }
+        else if (hasattr(obj, "_repr_json_"))
+        {
+            pub_data["application/json"] = static_cast<std::string>(
+                py::str(obj.attr("_repr_json_")())
+            );
+        }
+        else if (hasattr(obj, "_repr_jpeg_"))
```
Carreau: Most of these are usually not else-if. All the available mimetype of an object get computed by IPython.  You usually don't know which mimetype a frontend can display.

martinRenou: Ok, thanks!

src/xinterpreter.cpp
martinRenou: Maybe, is `_ipython_display_` something that is used a lot? Is it used in widget libraries? I wouldn't want to send a warning every single time the user wants to display widgets.

martinRenou: Ok I have my answer for widgets, I guess we can display a warning

martinRenou:
@SylvainCorlay Ok to squash and merge? I'll improve the global `display` function later.

PR #19, opened 2018-11-20
Clean display and logger

martinRenou:


src/xlogger.hpp
```diff
@@ -11,27 +11,21 @@
 #define XPYT_LOGGER_HPP
 
 #include <string>
-#include <vector>
-#include <functional>
 
 namespace xpyt
 {
     class xlogger
```
SylvainCorlay: I like these changes very much.

Quick item on `xlogger`: maybe rename this to `xstream` since it is not about error logging but standard output redirection.

For the `pipe_name` I would rather go with `stream_name`.

martinRenou: Ok :) I'll do those changes in a separate PR

martinRenou: `publisher`?

PR #21, opened 2018-11-21
Implement comms for widgets support

martinRenou:
- [x] Add support of widgets created from the back-end
- [x] Add support of widgets created from the front-end


src/xutils.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+    py::object cppmessage_to_pymessage(const xeus::xmessage& msg)
+    {
+        xeus::xjson json_msg;
+        json_msg["header"] = msg.header();
+        json_msg["parent_header"] = msg.parent_header();
+        json_msg["metadata"] = msg.metadata();
+        json_msg["content"] = msg.content();
+
+        py::dict py_msg = xjson_to_pyobj(json_msg);
```
SylvainCorlay: Doing

```cpp
py_msg["header"] = xjson_to_pyobj(msg.header());
py_msg["parent_header"] = xjson_to_pyobj(msg.parent_header());
py_msg["metadata"] = xjson_to_pyobj(msg.metadata());
py_msg["content"] = xjson_to_pyobj(msg.content());
```

Could prevent the copy into `json_msg`?

SylvainCorlay: Then we can work on a means to streamline nlohmann_json to python json conversion...

src/xutils.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+#include "pybind11/pybind11.h"
+
+namespace py = pybind11;
+
+namespace xpyt
+{
+
+    xeus::xjson pyobj_to_xjson(py::object obj);
+    py::object xjson_to_pyobj(const xeus::xjson& json);
```
SylvainCorlay: Nitpicking: how about

`json_to_py`
`py_to_json`

SylvainCorlay: Also, how about inluding `nlohmann/json.hpp` and dropping xjson wherever possible?

@JohanMabille

src/xutils.cpp
SylvainCorlay: Actually, (in a different PR), we could override `to_json` and `from_json` for `py::object`. Just like the JSON module of python, we would check if the object have `from_json` and `to_json` overloaded...

The only issue is that we would need to override these in the pybind11 namespace...

martinRenou:
```c++
py_msg["header"] = xjson_to_pyobj(msg.header());
py_msg["parent_header"] = xjson_to_pyobj(msg.parent_header());
py_msg["metadata"] = xjson_to_pyobj(msg.metadata());
py_msg["content"] = xjson_to_pyobj(msg.content());
```
will only transform one big copy into 4 smaller copies, but at the end of the day it will be the same, am I right?

src/xutils.hpp
martinRenou: If I do that, do you think I should define `xpyt::xjson`?

SylvainCorlay: I don't think so. The xjson header of xeus comes from our previous usage of rapidjson. We were adding utility functions.

src/xutils.cpp
SylvainCorlay: No. This would save a copy.

SylvainCorlay:
giphy
Noice.

PR #22, opened 2018-11-22
Rename logger to stream

martinRenou:


JohanMabille:
I'm not in favor of this change, this could be confusing with the notion of streams in the STL. Except if it is planned to implement all the stream operations in `xstream`, this class should have a different name.

SylvainCorlay:
As per in-person discussion, we agreed that "stream" was better than "logger" or "publisher".

If we add a prefix to classes exposed to python, we will do it for all at once.

PR #23, opened 2018-11-22
Fix compilation warnings

martinRenou:
See https://pybind11.readthedocs.io/en/stable/faq.html?highlight=greater%20visibility#someclass-declared-with-greater-visibility-than-the-type-of-its-field-someclass-member-wattributes


SylvainCorlay:
Did you run into the warning?

Quick note: it might be interesting to link to the section of the documentation in a comment.

martinRenou:
Yes I did. I agree, I'll add a comment

PR #26, opened 2018-11-27
Hack for matplotlib support

martinRenou:
`matplotlib` depends on `IPython`, which is annoying... We need to monkey patch some functions of `IPython` to make it work. This monkey patch should be removed soon, once this [PR](https://github.com/jupyter-widgets/ipywidgets/pull/2272) is merged in `ipywidgets`. At some point it would be nice to remove matplotlib's dependency for `IPython`


notebooks/xeus-python.svg
```diff
@@ -0,0 +1,94 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Generator: Adobe Illustrator 21.1.0, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+
+<svg
```
SylvainCorlay: GitHub cannot render SVGs in notebooks unfortunately. We need to upload a png for the notebook.

PR #29, opened 2018-11-28
Support for input requests

SylvainCorlay:
input


src/xstream.cpp
```diff
@@ -34,9 +34,28 @@ namespace xpyt
...
     PYBIND11_EMBEDDED_MODULE(xeus_python_stream, m) {
         py::class_<xstream>(m, "XPythonStream")
-            .def(py::init<std::string>())
-            .def("write", &xstream::write);
+            .def(py::init<std::string>());
+
+        m.def("write", write)
+         .def("input", input, py::arg("prompt") = "")
+         .def("raw_input", raw_input, py::arg("prompt") = "")
+         .def("getpass", getpass, py::arg("prompt") = "");
```
SylvainCorlay: Hesitant to put these into a `xeus_python_input` module.

SylvainCorlay: need to add an unused `stream` argument to `getpass`.

src/xinterpreter.cpp
```diff
@@ -106,6 +107,8 @@ namespace xpyt
             return kernel_res;
         }
 
+        forward_input(allow_stdin);
```
SylvainCorlay: I would like to make this a scope guard. For now, mimicking `ipykernel`.

src/xinterpreter.cpp
```diff
@@ -251,4 +255,43 @@ namespace xpyt
...
+            builtins.attr("input") = xeus_python_stream.attr("input");
+        }
+        py::module getpass = py::module::import("getpass");
+        m_sys_getpass = getpass.attr("getpass");
+        getpass.attr("getpass") = xeus_python_stream.attr("getpass");
+    }
+
+    void interpreter::restore_input()
+    {
+        if (PY_MAJOR_VERSION == 3)
```
SylvainCorlay: Could be a `#if` instead of `if`...

src/xstream.cpp
```diff
@@ -34,9 +34,28 @@ namespace xpyt
...
+    }
+
+    void raw_input(const std::string& prompt)
+    {
+        xeus::get_interpreter().input_request(prompt, false);
+    }
+
+    void getpass(const std::string& prompt)
+    {
+        xeus::get_interpreter().input_request(prompt, true);
```
SylvainCorlay: TODO: Await response and flush stdin/stdout buffers?

https://github.com/ipython/ipykernel/blob/118b684f75d51d72819ca90b69a245e9b02067a5/ipykernel/kernelbase.py#L875

src/xinterpreter.cpp
martinRenou: You should maybe call `forward_input` only if `allow_stdin` is true? And remove the argument for the `forward_input` method, as you don't use it.

src/xinterpreter.cpp
```diff
@@ -162,6 +165,7 @@ namespace xpyt
             kernel_res["evalue"] = evalue;
             kernel_res["traceback"] = traceback;
         }
+        restore_input();
```
martinRenou: Same for `restore_input`?

SylvainCorlay: Resolved, now using a scope guard.

src/xinterpreter.cpp
SylvainCorlay: Resolved as implemented in a scope guard.

src/xstream.cpp
SylvainCorlay: Closing and adding a comment in the thread.

src/xinterpreter.cpp
SylvainCorlay: Resolved.

src/xstream.cpp
SylvainCorlay: Done.

SylvainCorlay:
TODO: Await response and flush stdin/stdout buffers.

https://github.com/ipython/ipykernel/blob/118b684f75d51d72819ca90b69a245e9b02067a5/ipykernel/kernelbase.py#L875

SylvainCorlay:
Merging! cc @jcfr @ihnorton (cf screencast above).

src/xinput.cpp
```diff
@@ -0,0 +1,126 @@
...
+        py::module xeus_python_input = py::module::import("xeus_python_input");
+
+        // Forward input()
+        py::module builtins = py::module::import(XPYT_STRINGIFY(XEUS_PYTHON_BUILTINS));
+        m_sys_input = builtins.attr("input");
+        builtins.attr("input") = allow_stdin ? xeus_python_input.attr("input")
+                                             : xeus_python_input.attr("notimplemented");
+
+        // Forward raw_input()
+        if (PY_MAJOR_VERSION == 2)
```
martinRenou: Could it be a `#if PY_MAJOR_VERSION <= 2`?

src/xinput.cpp
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    input_redirection::~input_redirection()
+    {
+        // Restore input()
+        py::module builtins = py::module::import(XPYT_STRINGIFY(XEUS_PYTHON_BUILTINS));
+        builtins.attr("input") = m_sys_input;
+
+        // Restore raw_input()
+        if (PY_MAJOR_VERSION == 2)
```
martinRenou: `#if PY_MAJOR_VERSION <= 2`?

SylvainCorlay: I don't think we will ever support Python 1!

src/xinterpreter.cpp.orig
```diff
@@ -0,0 +1,392 @@
+/***************************************************************************
```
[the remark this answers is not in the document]
SylvainCorlay: Removing it on the master branch.

src/xcomm.cpp.orig
```diff
@@ -0,0 +1,166 @@
+/***************************************************************************
```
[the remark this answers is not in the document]
SylvainCorlay: idem

src/xinput.cpp
martinRenou: I'm not talking about Python1. What I meant is that we could check that at compilation time

SylvainCorlay: Yes, that is what I was doing earlier. Why not!

Issue #24, opened 2018-11-23
AttributeError for the logger


martinRenou:
I had this error:
```
AttributeError: 
'xeus_python_logger.XPythonLogger' object has no attribute 'flush'
```
I don't know how to reproduce it. Maybe we should implement `flush`, if it makes sense.

SylvainCorlay:
Yes I think that it makes sense to implement `flush`. You could also try out not sending the message until flush is called.

martinRenou:
According to the [documentation](https://docs.python.org/2/library/sys.html#sys.stderr):

>  stdout and stderr needn’t be built-in file objects: any object is acceptable as long as it has a write() method that takes a string argument.

Anyway, I think you're right and we should implement a flush method, because it's there in the original `stdout` and in `ipython`'s.

martinRenou:
I tried locally. I implemented a `flush` method, and I kept a buffer containing the messages while sending them only when `flush` is called. And apparently flush is not called. At least not when using the `print` function.

martinRenou:
This is fixed #89

PR #31, opened 2018-11-28
WIP - Implement flush method for stream

martinRenou:
Implementation of flush for #24


SylvainCorlay:
Note for future self: in core IPython, flush is implemented and called in the end of write.

PR #42, opened 2018-12-04
Implement fast json conversion

### LINKED ISSUE #25: Fast to_json and from_json
`to_json` and `from_json` for `py::object` should be optimized a bit. Instead of parsing string representation of the object we should implement recursive functions that create the `json`/`py::object`

martinRenou:
Fixes #25

### DIFF
--- a/src/xutils.cpp
+++ b/src/xutils.cpp
@@ -10,6 +10,7 @@
 #include <string>
 #include <vector>
 #include <stdexcept>
+#include <cmath>
 
 #include "nlohmann/json.hpp"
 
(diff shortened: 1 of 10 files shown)


src/xutils.cpp
```diff
@@ -83,20 +84,106 @@ namespace xpyt
...
+            }
+            if (py::isinstance<py::float_>(obj))
+            {
+                return obj.cast<double>();
+            }
+            if (py::isinstance<py::str>(obj))
+            {
+                return obj.cast<std::string>();
+            }
+            if (py::isinstance<py::tuple>(obj) || py::isinstance<py::list>(obj))
```
martinRenou: I need to add `py::set`

src/xutils.cpp
```diff
@@ -83,20 +84,106 @@ namespace xpyt
...
+            if (py::isinstance<py::dict>(obj))
+            {
+                json out;
+                for (py::handle key: obj)
+                {
+                    out[key.cast<std::string>()] = to_json_impl(obj[key]);
+                }
+                return out;
+            }
+            throw std::runtime_error("to_json not implemented for this type of object: " + obj.cast<std::string>());
```
martinRenou: If the `Python` object is iterable I should be able to json-ify it

src/xutils.cpp
```diff
@@ -83,20 +84,106 @@ namespace xpyt
...
+            if (j.is_boolean())
+            {
+                return py::bool_(j.get<bool>());
+            }
+            if (j.is_number())
+            {
+                double number = j.get<double>();
+                if (number == std::floor(number))
+                {
+                    return py::int_(j.get<int>());
```
ihnorton: Could use a note that this is `int` everywhere rather than `size_t` because of JSON representation limitation (IIUC).

SylvainCorlay:
Corrected a typo in appveyor. Feel free to squash the commit when you rebase.

(this now needs a rebase with the merging of input requests.)

src/xutils.cpp
SylvainCorlay: JSON and JS can only represent floating points, no real integer. The choice to use an integer when the value is integral is to be consistent with the behavior of the JSON module of Python.

martinRenou: That's the behavior of the JSON module of Python. If the number value in the JSON can be represented as an integer value it creates an integer, not a floating point number. But the question is, is the `py::int_` an `int`, `long`, `long long` or `ptrdiff_t`?

martinRenou: I guess it depends on the value itself... https://github.com/pybind/pybind11/blob/master/include/pybind11/pytypes.h#L1076

PR #44, opened 2018-12-09
WIP - Update display

### LINKED ISSUE #20: Implement update_display

SylvainCorlay:
Fixes #20.


src/xdisplay.cpp
```diff
@@ -13,61 +13,18 @@
...
-#include "xutils.hpp"
 #include "xdisplay.hpp"
+#include "xutils.hpp"
 
 namespace py = pybind11;
 namespace nl = nlohmann;
 
 namespace xpyt
 {
-    xdisplayhook::xdisplayhook() : m_execution_count(0)
```
SylvainCorlay: called it `mime_bundle_repr` so that it may as well be the standard `mime_bundle_repr` in xeus-cling for `pybind11::object`.

src/xdisplay.cpp
```diff
@@ -125,11 +82,110 @@ namespace xpyt
...
+                nl::json::object()
+            );
+        }
+    }
+
+    /***************************
+     * xdisplay implementation *
+     ***************************/
+
+    void xdisplay(const py::object& obj, const py::object display_id, bool update)
```
SylvainCorlay: The display function differs from the display hook in that it does not use `publish_execution_result` but `display_data` and `update_display_data`, and therefore does not produce an output prompt.

PR #50, opened 2018-12-13
Update xeus version

martinRenou:


SylvainCorlay:
Similarly to xeus-cling, if you update to 0.18.1 and use `blocking_input_request`, it should also remove some code!

PR #60, opened 2018-12-18
Clean code inspection

martinRenou:
- remove token_at_cursor function copied from IPython
- implement code inspection using only jedi


martinRenou:
Before:
before_inspect

After:
after_inspect

PR #66, opened 2019-01-02
Replace embedded modules by simple modules

martinRenou:


martinRenou:
I think this PR won't be merged, we'll simply remove modules one by one, it will be less code, less complexity.

PR #75, opened 2019-01-04
Add Python 2 support

martinRenou:


martinRenou:
Python2 support works for me locally, I'll try adding Python tests in this PR

PR #93, opened 2019-04-03
WIP - First debugger prototype

martinRenou:


martinRenou:
This does not do anything useful yet.
It only spawns ptvsd, and creates a socket for communication with it.
The communication will start only if:
- the global `debugging()` Python function has been called in the Notebook
- A client (e.g. UI component in JupyterLab) has opened a `jupyter.debugger` Comm

martinRenou:
Note to self: Do not forget to add ptvsd as dependency

martinRenou:
@JohanMabille Closing in favor of your work in #107, #108

PR #96, opened 2019-04-08
Fix multithreading issue

### LINKED ISSUE #95: Output widget + multithread weirdness
When using output widgets and adding data from multiple threads, xeus-python doesn't seem to properly update the contents of the output widget. Only upon triggering the kernel from the frontend (e.g. by activating auto-completion or executing a `print` statement) the contents of the output widget get updated.

I am using the following output redirector:

```py
import threading
import ipywidgets as widgets
import sys

output_registry = {}

class OutputRedirector:
    def __init__(self, original):
        self.original = original
    
    def write(self, msg):
        thread_name = threading.currentThread().name
        if thread_name != 'MainThread':
            output_registry[threading.currentThread().getName()].append_stdout(msg)
        else:
            self.original.write(msg)

    def flush(self):
        self.original.flush()

    # necessary for ipython, but **not** for xeus-python!
    def set_parent(self, parent):
        self.original.set_parent(parent)

sys.stdout = OutputRedirector(sys.stdout)
```

And now you can create a new thread with outptu widget, e.g. 

```py
def func():
    while True:
        print("Hello?")

def new_thread():
    output_registry['test'] = widgets.Output()
    thread = threading.Thread(target=func, name="test")
    thread.start()
    return output_registry['test']
```

martinRenou:
Should fix #95

### DIFF
--- a/include/xeus-python/xinterpreter.hpp
+++ b/include/xeus-python/xinterpreter.hpp
@@ -16,6 +16,7 @@
 #endif
 
 #include <string>
+#include <memory>
 #include <vector>
 
 #include "nlohmann/json.hpp"
@@ -34,6 +35,8 @@ namespace xpyt
     class XEUS_PYTHON_API interpreter : public xeus::xinterpreter
     {
     public:
+
+        using gil_scoped_release_ptr = std::unique_ptr<py::gil_scoped_release>;
 
         interpreter(int argc, const char* const* argv);
         virtual ~interpreter();
@@ -72,7 +75,7 @@ namespace xpyt
         // interpreter wants to execute Python code. This means that whenever
         // the interpreter will execute Python code it will need to create an
         // `gil_scoped_acquire` instance first.
-        py::gil_scoped_release* m_release_gil;
+        gil_scoped_release_ptr m_release_gil = nullptr;
     };
 }
 
--- a/include/xeus-python/xinterpreter.hpp
+++ b/include/xeus-python/xinterpreter.hpp
@@ -72,9 +72,7 @@ namespace xpyt
         // interpreter wants to execute Python code. This means that whenever
         // the interpreter will execute Python code it will need to create an
         // `gil_scoped_acquire` instance first.
-        // This member must be stay the last one, so that other Python object
-        // members can be initialized without crashing.
-        py::gil_scoped_release m_release_gil;
+        py::gil_scoped_release* m_release_gil;
     };
 }
 
--- a/include/xeus-python/xinterpreter.hpp
+++ b/include/xeus-python/xinterpreter.hpp
@@ -66,6 +66,15 @@ namespace xpyt
 
         py::object m_displayhook;
         std::vector<std::string> m_inputs;
+
+        // The interpreter has the same scope as a `gil_scoped_release` instance
+        // so that the GIL is not held by default, it will only be held when the
+        // interpreter wants to execute Python code. This means that whenever
+        // the interpreter will execute Python code it will need to create an
+        // `gil_scoped_acquire` instance first.
+        // This member must be stay the last one, so that other Python object
+        // members can be initialized without crashing.
+        py::gil_scoped_release m_release_gil;
     };
 }
 
--- a/src/xinterpreter.cpp
+++ b/src/xinterpreter.cpp
@@ -8,6 +8,7 @@
 ****************************************************************************/
 
 #include <algorithm>
+#include <memory>
 #include <string>
 #include <utility>
 #include <vector>
@@ -38,7 +39,9 @@ namespace xpyt
 {
     void interpreter::configure_impl()
     {
-        m_release_gil = new py::gil_scoped_release();
+        // The GIL is not held by default by the interpreter, so every time we need to execute Python code we
+        // will need to acquire the GIL
+        m_release_gil = gil_scoped_release_ptr(new py::gil_scoped_release());
 
         py::gil_scoped_acquire acquire;
         py::module jedi = py::module::import("jedi");
--- a/src/xinterpreter.cpp
+++ b/src/xinterpreter.cpp
@@ -38,6 +38,8 @@ namespace xpyt
 {
     void interpreter::configure_impl()
     {
+        m_release_gil = new py::gil_scoped_release();
+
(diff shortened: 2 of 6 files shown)


src/xserver.cpp
```diff
@@ -0,0 +1,77 @@
...
+            {
+                zmq::multipart_t wire_msg;
+                wire_msg.recv(m_shell);
+                xserver::notify_shell_listener(wire_msg);
+            }
+        }
+        else
+        {
+            // Give some time for Python threads to resume their job
+            py::module::import("time").attr("sleep")(0.02);
```
martinRenou: I'm not sure if `20ms` is fine

martinRenou: `20ms` seems fine. I tested running both a Python thread on the background and a slider widget communication, even when moving the slider the thread keeps running.

wolfv:
great! Looks like a clean solution!

martinRenou:
Please do not merge, I have a cleaner solution

src/xserver.cpp
martinRenou: This is not relevant anymore, as we are not sleeping

src/xserver.cpp
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    void server::poll_msg()
+    {
+        zmq::pollitem_t items[]
+            = { { m_controller, 0, ZMQ_POLLIN, 0 }, { m_shell, 0, ZMQ_POLLIN, 0 } };
+
+        py::gil_scoped_release release;
+        zmq::poll(&items[0], 2, -1);
+        py::gil_scoped_acquire acquire;
```
martinRenou: I am now releasing the GIL when waiting for new messages, allowing Python secondary threads to resume their job

JohanMabille: `gil_scoped_release` and `gil_scoped_acquire` are scoped guard, meaning they are meant to be used in a scope, their destructor will do the opposite operation of the one in their constructor when the thread exists the scope. So the code should be
```cpp
{
    py::gil_scoped_release release;
    zmq::poll(&items[0], 2, -1);
}
```

wolfv: @martinRenou why don't you remove all of the xserver here? You can release the GIL after you finish the execute_request and acquire the GIL when you get the next execute request. This should be only a 2 line change, no?

martinRenou: Yes I will. Did not have time to do it yet.

martinRenou: I need to put a `py::gil_scoped_acquire` for every request where I execute Python code. But I also need to check if the acquire works like a reentrant mutex, if it's the case we need to be careful that the GIL was not acquired earlier.

martinRenou: Mmmmh... It is not as simple as I thought it would be. The acquire crashes.
```
Program received signal SIGSEGV, Segmentation fault.
x00007ffff75cab80 in threadstate_getframe ()

traceback:
#0  0x00007ffff75cab80 in threadstate_getframe () from /home/martin/miniconda3/envs/xeus-python/lib/libpython3.7m.so.1.0
#1  0x00007ffff7658caa in PyEval_GetBuiltins () from /home/martin/miniconda3/envs/xeus-python/lib/libpython3.7m.so.1.0
#2  0x00007ffff77afa25 in pybind11::detail::get_internals() () from /home/martin/miniconda3/envs/xeus-python/lib/libxeus-python.so.0
#3  0x00007ffff77b0539 in pybind11::gil_scoped_acquire::gil_scoped_acquire() [clone .lto_priv.253] () from /home/martin/miniconda3/envs/xeus-python/lib/libxeus-python.so.0
```

martinRenou: Ok it's fixed now

martinRenou:
This breaks widgets support. I guess there is a missing acquire in the Comms implementation.

martinRenou:
We will need to be really careful now concerning the GIL thing.

martinRenou:
So, I came up with this solution: 
- The GIL is not held by default, allowing other threads to execute Python code if needed.
- Whenever the interpreter needs to execute Python code it must hold the GIL. There is two ways to acquire the GIL:
  - Using a `py::gil_scoped_acquire` instance (unsafe, we need to be sure that the GIL is not already held by the current thread when creating the instance).
  - Using the `XPYT_HOLDING_GIL` macro that I created (safe, it first checks if the GIL is already held, if not it creates a `py::gil_scoped_acquire` instance). This macro takes a function call/code as argument.

include/xeus-python/xinterpreter.hpp
```diff
@@ -66,6 +66,15 @@ namespace xpyt
...
         std::vector<std::string> m_inputs;
+
+        // The interpreter has the same scope as a `gil_scoped_release` instance
+        // so that the GIL is not held by default, it will only be held when the
+        // interpreter wants to execute Python code. This means that whenever
+        // the interpreter will execute Python code it will need to create an
+        // `gil_scoped_acquire` instance first.
+        // This member must be stay the last one, so that other Python object
+        // members can be initialized without crashing.
+        py::gil_scoped_release m_release_gil;
```
martinRenou: This change makes it impossible to inherit from `xinterpreter` while adding a Python object as a member...

martinRenou: This should be resolved now

martinRenou:
@JohanMabille Could you tell me what you think of the last commit? I suppose this solution works. Now there is no problem inheriting from the `xinterpreter` anymore (we could not inherit from it while having `py::object` members before, due to the fact that we were not holding the GIL at construction). So now the GIL is held at the interpreter construction, but we release it in `configure_impl`.

src/xinterpreter.cpp
```diff
@@ -38,6 +38,9 @@ namespace xpyt
 {
     void interpreter::configure_impl()
     {
+        m_release_gil = new py::gil_scoped_release();
```
JohanMabille: I think this deserves a comment (having a scope object dynamically instantiated is weird at first sight) and a delete in the destructor of the interpreter.

martinRenou: Made it an `std::unique_ptr` as discussed in person :)

PR #102, opened 2019-05-06
Test Windows CI

martinRenou:


martinRenou:
The failing build is more likely to be a dependency issue

PR #109, opened 2019-06-27
Upgraded to xeus 0.20.0

JohanMabille:


JohanMabille:
xeus 0.20.0 is not available on all mirrors yet. I will retrigger the build later

PR #112, opened 2019-07-01
Added first test of debugger

JohanMabille:


test/test_debugger.cpp
```diff
@@ -0,0 +1,167 @@
...
+}
+
+nl::json make_init_request()
+{
+    nl::json req = {
+        {"type", "request"},
+        {"seq", 1},
+        {"command", "initialize"},
+        {"arguments", {
+            {"cliendID", "vscode"},
```
wolfv: This looks like a typo

JohanMabille: Well this is just a cliendID example in the tests, I don't think it really matters.

PR #117, opened 2019-07-25
Implemented breakpoints in cells

JohanMabille:
This should be the last PR before releasing ;)

fc


SylvainCorlay:
@jasongrout will appreciate the gif.

PR #127, opened 2019-08-22
Accept None as input in xutils::pylist_to_zmq_buffers

martinRenou:
Partially fixing #124


martinRenou:
There seem to be an issue, the CI was not triggered. And I cannot push more commits to this PR...

PR #130, opened 2019-08-26
WIP - MacOS issue

martinRenou:


martinRenou:
This can be closed, there was actually no MacOS issue, the user was running an old version of xeus-python

PR #136, opened 2019-09-27
Always save code to tmp files

### LINKED ISSUE #135: inspect.getsource does not work
```python
import inspect

def foo():
    pass

inspect.getsource(foo)
``` 
does not seem to work in the xpython kernel... 

```
Line 6:   inspect.getsource(foo)

File /home/martin/miniconda3/envs/foo/lib/python3.7/inspect.py, in getsource:
Line 973:   lines, lnum = getsourcelines(object)

File /home/martin/miniconda3/envs/foo/lib/python3.7/inspect.py, in getsourcelines:
Line 955:   lines, lnum = findsource(object)

File /home/martin/miniconda3/envs/foo/lib/python3.7/inspect.py, in findsource:
Line 786:   raise OSError('could not get source code')

OSError: could not get source code
```

Resulting in https://github.com/QuantStack/py2vega/ to not work (**EDIT:** at least not in the case Python functions, it would work with string functions)

martinRenou:
This will fix #135. It allows the inspect module to retrieve the Python source code.

working

### DIFF
--- a/src/xdebugger.cpp
+++ b/src/xdebugger.cpp
@@ -126,9 +126,11 @@ namespace xpyt
         std::string next_file_name = xeus::get_cell_tmp_file(get_tmp_prefix(), next_id, ".py");
         std::clog << "debugger filename = " << next_file_name << std::endl;
 
-        std::ofstream out(next_file_name);
+        std::ofstream out;
+        out.open(next_file_name, std::ofstream::out | std::ofstream::trunc);
         out << code << std::endl;
-        
+        out.close();
+
         nl::json reply = {
             {"type", "response"},
             {"request_seq", message["seq"]},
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
Does ipykernel need that to achieve the same thing?

martinRenou:
I don't know, I can check. I only checked the inspect module implementation.

martinRenou:
File names in ipykernel are in the form of `<ipython-input-1-9bc52a56589c>`, they don't look like Python file names.
In xeus-python file names are in the form of `/tmp/xpython_6019/[1].py`, so my guess is that the inspect module looks for this file and does not find it.

martinRenou:
Still investigating what ipykernel/ipython does. They might use different compiling options.

martinRenou:
So it seems that IPython is monkey patching the builtin `linecache` module: https://github.com/ipython/ipython/blob/56ea43f4378f4a87e3aec4cfae4dca26bb69c2ec/IPython/core/compilerop.py

And the `linecache` module that is used by the `inspect` module for retrieving the source code: https://github.com/python/cpython/blob/3.7/Lib/inspect.py#L784

I wonder what's best. I feel uncomfortable monkey patching a builtin module.

martinRenou:
As per in-person discussion. We'll go for the IPython solution.

jtpio:
The main reason being the number of files created?

martinRenou:
Yes. We might come back on this decision if we take the decision of always starting the debugger (because starting the debugger results in saving cell code in files on disk).

PR #145, opened 2019-10-18
Switched to xtl murmur2 implementation

JohanMabille:


src/xutils.hpp
```diff
@@ -56,6 +56,7 @@ namespace xpyt
 
     void exec(const py::object& code, const py::object& scope = py::globals());
 
+    size_t get_hash_seed();
```
SylvainCorlay: std::size_t?

JohanMabille: Ah good catch. I'll fix it in the next PR.

PR #168, opened 2019-11-15
Add source request

jtpio:
The idea is to implement the `source` request in `xeus-python`, so we are able to retrieve the source code for a give `source.path`.

DAP: https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Source

Related:

- https://github.com/jupyterlab/debugger/pull/170
- https://github.com/jupyterlab/debugger/issues/169

### DIFF
--- a/test/test_debugger.cpp
+++ b/test/test_debugger.cpp
@@ -239,6 +239,21 @@ nl::json make_dump_cell_request(int seq, const std::string& code)
     return req;
 }
 
+nl::json make_source_request(int seq, const std::string& path)
+{
+    nl::json req = {
+        {"type", "request"},
+        {"seq", seq},
+        {"command", "source"},
+        {"arguments", {
+            {"source", {
+                {"path", path}
+            }}
+        }}
+    };
+    return req;
+}
+
 nl::json make_stepin_request(int seq, int thread_id)
 {
     nl::json req = {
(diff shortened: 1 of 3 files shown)


test/test_debugger.cpp
```diff
@@ -403,6 +419,23 @@ bool debugger_client::test_set_breakpoints()
...
+    std::string code = make_code();
+    m_client.send_on_control("debug_request", make_dump_cell_request(4, code));
+    nl::json reply = m_client.receive_on_control();
+    std::string path = reply["content"]["body"]["sourcePath"].get<std::string>();
+
+    m_client.send_on_control("debug_request", make_source_request(5, path));
+    nl::json rep = m_client.receive_on_control();
+
+    nl::json source = rep["content"]["body"]["content"];
+    bool res = source == code + '\n';
```
jtpio: Do we need to add an extra `\n` in the `dumpCell` request?

https://github.com/QuantStack/xeus-python/blob/8faad74899564f267ef670f944f2285eb53e214f/src/xdebugger.cpp#L154

jtpio:
cc @JohanMabille 

Feel free to push to that branch directly in case we decide to drop the extra `std::endl`:

https://github.com/QuantStack/xeus-python/blob/8faad74899564f267ef670f944f2285eb53e214f/src/xdebugger.cpp#L154

JohanMabille:
I prefer to merge it as it is now (I'm starting again the tests on Windows because the failure is not related) and refactor after?

jtpio:
Sure, sounds good :+1:

PR #183, opened 2019-12-12
Make pybind11 not part of xeus-python's interface

SylvainCorlay:


CMakeLists.txt.orig
```diff
@@ -0,0 +1,271 @@
+############################################################################
+# Copyright (c) 2016, Martin Renou, Johan Mabille, Sylvain Corlay and      #
```
JohanMabille: I think you don't want to commit this file.

PR #186, opened 2019-12-18
Deadlock

JohanMabille:


SylvainCorlay:
It seems that we are still getting these deadlocks in master after this PR unfortunately.

Issue #190, opened 2020-01-09
Multiple users running xeus-python with a debugger fails.


yulius-fxpal:
When xeus-python is used with Jupyter Hub and the Jupyter Lab debugger and multiple people run xeus-python kernel, you will get this error.

* User A will start xeus-python and create `/tmp/xpython_debug_logs` owned by User A.
* User B will start xeus-python and try to connect to ptvsd with `/tmp/xpython_debug_logs` but won't be able to create new log files.

```
Jan  9 12:21:36 elk04 jupyterhub[50627]: Exception raised when trying to import ptvsd                                   Jan  9 12:21:36 elk04 jupyterhub[50627]: import ptvsd                                                                   
Jan  9 12:21:36 elk04 jupyterhub[50627]: ptvsd.enable_attach(('127.0.0.1',5854), log_dir='/tmp/xpython_debug_logs')
Jan  9 12:21:36 elk04 jupyterhub[50627]: PermissionError - [Errno 13] Permission denied: '/tmp/xpython_debug_logs/ptvsd-50904.log'`
```

The fix I propose is to append the username or process id to `xpython_debug_logs`.

https://github.com/jupyter-xeus/xeus-python/blob/ec7338cafbe0f172bf9f1dcd0c0088713b35d167/src/xdebugger.cpp#L323

JohanMabille:
That's a good idea. Would you like to open a PR? `xeus` provides the `get_current_pid` function in `xsystem.hpp` so it should be easy to fix.

yulius-fxpal:
@JohanMabille just created a PR.  Hopefully it will go into the next release.  Thanks.

jtpio:
This should now be fixed in the latest `0.6.9` release.

Were you able to test with the new version @yulius-fxpal?

yulius-fxpal:
@jtpio Sorry for the delay.  Yes, it fixes the issue.  Thank you!  Shall I close the bug?

PR #191, opened 2020-01-09
#190 bug fix: append pid to xpython_debug_logs directory

yulius-fxpal:
Append pid to xpython_debug_logs directory name.  Fix for bug #190


JohanMabille:
Thanks!

PR #193, opened 2020-01-22
fix wrongly incremented cursor position in inspect

### LINKED ISSUE #192: Wrong inspect cursor position
If we have a cell `from plotly import express` 
When the cursor `|` is on `from plotly| import express`, the docstring of plotly is not shown.
Instead, if the cursor is on `from plotl|y import express`, the docstring is shown.

cnydw:
Fixes #192


martinRenou:
This looks good to me :) Thanks a lot for fixing this

PR #195, opened 2020-01-30
Follow the DAP for inspectVariables response

### LINKED ISSUE #194: Follow the DAP for the inspectVariables response
At the moment the `inspectVariables` request gets all the variables from the running kernel with `globals()`, which are then returned as-is in the response.

Ideally, the response could follow the same structure as for the `variables` response from the DAP, and return a list of [`Variable`](https://microsoft.github.io/debug-adapter-protocol/specification#Types_Variable) (probably with the optional fields left out, at least for now).

Related to https://github.com/jupyterlab/debugger/issues/118.

jtpio:
Fixes #194.

Example of a `inspectVariables` response:

```json
{
  "type": "response",
  "success": true,
  "request_seq": 10,
  "command": "inspectVariables",
  "body": {
    "variables": [
      {
        "name": "__builtins__",
        "value": "<module 'builtins' (built-in)>",
        "variablesReference": 0
      },
      {
        "name": "display",
        "value": "<built-in method display of PyCapsule object at 0x7f0952358f00>",
        "variablesReference": 0
      },
      {
        "name": "i",
        "value": 4,
        "variablesReference": 0
      },
      {
        "name": "j",
        "value": 8,
        "variablesReference": 0
      },
      {
        "name": "k",
        "value": 5,
        "variablesReference": 0
      }
    ]
  }
}
```

### TODO

- [x] Return a list of `Variable` for the `inspectVariables` response
- [x] Make sure to include other required fields
- [x] Improve tests for `inspectVariables`

### DIFF
--- a/test/test_debugger.cpp
+++ b/test/test_debugger.cpp
@@ -565,7 +565,7 @@ bool debugger_client::test_stack_trace()
 
     m_client.send_on_shell("execute_request", make_execute_request(make_code()));
     nl::json ev = m_client.wait_for_debug_event("stopped");
-    
+
     int seq = 6;
     m_client.send_on_control("debug_request", make_stacktrace_request(seq, 1));
     ++seq;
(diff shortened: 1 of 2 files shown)


test/test_debugger.cpp
```diff
@@ -632,8 +632,14 @@ bool debugger_client::test_inspect_variables()
...
     nl::json rep = m_client.receive_on_control();
 
-    nl::json var = rep["content"]["body"]["variables"];
-    bool res = var["i"] == 4 && var["j"] == 8 && var["k"] == 5;
+    nl::json vars = rep["content"]["body"]["variables"];
+    std::cout << vars << std::endl;
+    std::cout << vars.size() << std::endl;
+    size_t size = vars.size();
+    bool res = size >= 3;
+    res = res && vars[size-3]["name"] == "i" && vars[size-3]["value"] == 4;
```
jtpio: This should be improved by searching for the variable in the array of variables, instead of relying on a specific order.

jtpio:
cc @JohanMabille if you want to have a look

PR #199, opened 2020-02-17
Link with dl library when available

JohanMabille:


SylvainCorlay:
Do you know why we don't *always* need that?

JohanMabille:
We don't need to explicitely link with libdl since linking with libpython.so pulls libdl. However when only the static version of libpython is available, this is required. Since linking twice with the same dynamic library is not a problem, I don't see why we should test whether we link with the static or the dynamic libpython.

PR #201, opened 2020-02-28
Add warning when building as a PyPI wheel

SylvainCorlay:
The problem is that the banner is only visible in the console.


SylvainCorlay:
This may be improved if we provide a means to display the banner in the notebook interface of JupyterLab.

https://github.com/jupyterlab/jupyterlab/issues/7948

PR #208, opened 2020-03-13
Static build

JohanMabille:
@SylvainCorlay this is far from being finished and will need some refactoring when we have something that works.

For now I'm stucked with the link of xpython with libxeus.a.


JohanMabille:
Let's refactor after we upgrade to xeus 0.23.8.

The static build works when building both xeus and xeus-python on the same machine, so this issue is not a blocker for the wheel.

PR #211, opened 2020-03-17
Add `help_links` to `kernel_info_reply`

jtpio:
Add `help_links` to `kernel_info_reply` so the xeus-python documentation can be displayed by the frontend:

help-links

### DIFF
--- a/src/xinterpreter.cpp
+++ b/src/xinterpreter.cpp
@@ -291,6 +291,13 @@ namespace xpyt
         result["language_info"]["version"] = PY_VERSION;
         result["language_info"]["mimetype"] = "text/x-python";
         result["language_info"]["file_extension"] = ".py";
+
+        result["help_links"] = nl::json::array();
+        result["help_links"][0] = nl::json::object({
+            {"text", "Xeus-Python Reference"},
+            {"url", "https://xeus-python.readthedocs.io"}
+        });
+
         result["status"] = "ok";
         return result;
     }


src/xinterpreter.cpp
```diff
@@ -291,6 +291,13 @@ namespace xpyt
         result["language_info"]["version"] = PY_VERSION;
         result["language_info"]["mimetype"] = "text/x-python";
         result["language_info"]["file_extension"] = ".py";
+
+        result["help_links"] = nl::json::array();
+        result["help_links"][0] = nl::json::object({
+            {"text", "xeus-python Reference"},
```
SylvainCorlay: Should we capitalize Reference?

SylvainCorlay: Or Xeus...

jtpio: How about "Xeus-Python Reference"?

Reference is capitalized to stay consistent with `ipykernel`: https://github.com/ipython/ipykernel/blob/92f979eda8280a8898d5235c35df62ac3ee15c79/ipykernel/ipkernel.py#L88-L117

SylvainCorlay: Sounds good.

jtpio: Done :+1:

SylvainCorlay:
Thanks!

SylvainCorlay:
Would be really nice to have the same in the xeus-cling kernel :)

jtpio:
> Would be really nice to have the same in the xeus-cling kernel :)

Definitely. I'll open a similar PR later during the day.

JohanMabille:
Now we need to really improve the documentation ;)

PR #214, opened 2020-03-24
Added handle for dumping backtrace on segfault

JohanMabille:


src/main.cpp
```diff
@@ -25,6 +33,21 @@
 
 #include "xpythonhome.hpp"
 
+#ifdef __GNUC__
+void handler(int sig)
+{
+    void *array[10];
```
JohanMabille: I'll fix this in the next PR

PR #216, opened 2020-03-24
WIP: Allow specifying the sys path at compile time

SylvainCorlay:


src/xsyspath.cpp
```diff
@@ -0,0 +1,38 @@
...
+#if defined(XEUS_PYTHON_SYSPATH)
+        static const std::string syspath = XPYT_STRINGIFY(XEUS_PYTHON_SYSPATH);
+        // TODO: prefix each item in the ';' or ':' separated list with prefix_path()
+#if PY_MAJOR_VERSION == 2
+        Py_SetPath(const_cast<char*>(syspath.c_str()));
+#else
+        static const std::wstring wstr(syspath.cbegin(), syspath.cend());;
+        Py_SetPath(const_cast<wchar_t*>(wstr.c_str()));
+#endif
+#endif
```
SylvainCorlay: As per the `TODO:` above, I presume we would specify the paths in `XEUS_PYTHON_SYSPATH` in a relative fashion with respect to the installation prefix.

SylvainCorlay: Maybe the way to go would be to do a replacement of a magic string, like `SYS_PREFIX`.

SylvainCorlay:
This seems to throw a `std::regex_error`.

PR #220, opened 2020-03-25
Removed useless flags in static_dependencies build

JohanMabille:
Reverts #218


SylvainCorlay:
Note that #218 could be applied to core xeus, when building the `.so`. I think that it would make sense.

JohanMabille:
We don't want the static biuld of xpython to depend on libxeus.so, otherwise this latter would be installed by the wheel and could be overwrite an installation of xeus-cling for instance (or any other kernel built upon xeus).

SylvainCorlay:
> We don't want the static biuld of xpython to depend on libxeus.so, otherwise this latter would be installed by the wheel and could be overwrite an installation of xeus-cling for instance (or any other kernel built upon xeus).

I agree, but I think it makes sense for people who link with `libxeus.so` in general.

JohanMabille:
WIndows is not affected by this change, no need to wait for appveyor

PR #222, opened 2020-03-27
Fix missing arguments for mime_bundle_repr

### LINKED ISSUE #221: xpython does not pass the required include and exclude arguments to _repr_mimebundle_
when plotting an altair graph in a jupyter lab note book, I got the error:

TypeError: repr_mimebundle() missing 2 required positional arguments: 'include' and 'exclude'

I raised this with the altair people eg. @jakevdp and they thought it best that I raise it here.

Here is what I tried:
```
import altair as alt
alt.Chart(pd.DataFrame({'a': [1, 2, 3], 'b':[4, 0, 9]})).mark_point().encode(
    x='a:Q',
    y='b:Q')
```
and here is the complete error message: 

```
TypeError Traceback (most recent call last)
In [11]:
Line 1: points

TypeError: repr_mimebundle() missing 2 required positional arguments: 'include' and 'exclude'
```
here is my config :
XEUS 0.23.9
jupyter core : 4.6.3
jupyter-notebook : 6.0.3
qtconsole : not installed
ipython : 7.13.0
ipykernel : 5.2.0
jupyter client : 6.1.1
jupyter lab : 2.0.1
nbconvert : 5.6.1
ipywidgets : not installed
nbformat : 5.0.4
traitlets : 4.3.3
altair. : 4.0.1

martinRenou:
Fixes #221

IPython also calls `_repr_mimebundle_` with the `include` and `exclude` parameters, see https://github.com/ipython/ipython/blob/c2a558301946b6253ff3d0dd56f12585783258f7/IPython/core/formatters.py#L970.

I suspect this is a backward-incompatible change (our example Notebook does not work anymore).

### DIFF
--- a/notebooks/xeus-python.ipynb
+++ b/notebooks/xeus-python.ipynb
@@ -244,7 +244,7 @@
     "        self.address = address\n",
     "        self.picture = picture\n",
     "\n",
-    "    def _repr_mimebundle_(self):\n",
+    "    def _repr_mimebundle_(self, include=None, exclude=None):\n",
     "        return {\n",
     "            \"text/html\": \"\"\"<img src=\"{}\">\n",
     "                  <div><i class='fa-user fa'></i>: {}</div>\n",
--- a/src/xdisplay.cpp
+++ b/src/xdisplay.cpp
@@ -27,7 +27,7 @@ namespace nl = nlohmann;
 
 namespace xpyt
 {
-    nl::json mime_bundle_repr(const py::object& obj)
+    nl::json mime_bundle_repr(const py::object& obj, const py::object& include = py::none(), const py::object& exclude = py::none())
     {
         py::module py_json = py::module::import("json");
         py::module builtins = py::module::import(XPYT_BUILTINS);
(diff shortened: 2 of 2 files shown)


martinRenou:
Those parameters should also be added to the `display` function. Otherwise, they are useless.

SylvainCorlay:
This looks good to me.

SylvainCorlay:
Can you confirm this works with Altair?

martinRenou:
Yep
altair

SylvainCorlay:
Great!

martinRenou:
We still don't get reports from TravisCI, it's been 4 days since it does not work.

The AppVeyor build issue seems unrelated.

JohanMabille:
See [this incident](https://www.traviscistatus.com/incidents/rx6fhs3wqcln?u=49rdxf5zfpg9). Apparently it is not totaly solved. Besides, I cannot restart the builds anymore and it seems that my github account is not recognized correctly in travis.

I think we should remove the hook and move to Github apps, this should fix the issue. I will make a pass on all the repositories this afternoon.

JohanMabille:
The issue on appveyor is due to the client waiting forever a response from the kernel (his is related to the debugger). We see more and more often this issue, so we should investigate at some point.

SylvainCorlay:
> We still don't get reports from TravisCI, it's been 4 days since it does not work.

You can still see the builds here: https://travis-ci.org/github/jupyter-xeus/xeus-python

JohanMabille:
The error on travis-ci is the same as that of appveyor, let's merge!

PR #227, opened 2020-03-28
Change cmake build flag for splitting into a shared object and an executable

SylvainCorlay:
I hesitant to call it `XPYT_BUILD_SHARED` instead.

Thoughts?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,17 +18,19 @@ matrix:
         apt:
           sources:
             - ubuntu-toolchain-r-test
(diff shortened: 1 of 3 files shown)


SylvainCorlay:
@JohanMabille I am only getting the error on Travis.

Putting the build case in `allow_failures` as previously.

JohanMabille:
Ok, let's wait for the investigation tools in xeus, they should help us investigate this issue.

PR #232, opened 2020-03-30
Simplify cmake even more

SylvainCorlay:


CMakeLists.txt
```diff
@@ -286,12 +286,7 @@ endif ()
 # xpython
 # =======
 
-if (XPYT_SPLIT_XPYTHON)
-    add_executable(xpython ${XPYTHON_SRC})
-else ()
-    add_executable(xpython ${XPYTHON_SRC} ${XEUS_PYTHON_SRC} ${XEUS_PYTHON_HEADERS})
```
JohanMabille: We should keep this otherwise CMake won't rebuild `xpython` if one of this file changes locally.

JohanMabille: I missed the dependency on libxeus-python

PR #236, opened 2020-03-31
Matplotlib inline support

martinRenou:


src/xmatplotlib.cpp
```diff
@@ -0,0 +1,255 @@
...
+                    ip.showtraceback()
+                    return
+    finally:
+        show._to_draw = []
+        show._draw_called = False
+
+
+FigureCanvas = FigureCanvasAgg
+
+# def _enable_matplotlib_integration():
```
martinRenou: uncomment this

JohanMabille:
I'm really not in favor of this. This (support of matplotlib) should be done in a dedicated library or plugin that is eventually loaded by xeus-python.

AntoineToussaint:
Would we need something similar to support plotly?

martinRenou:
@AntoineToussaint could you open a specific issue for plotly? If you see an error with plotly, we might be able to fix it cleanly. Matplotlib is a bit more difficult.

AntoineToussaint:
Oh, yes, of course.

martinRenou:
@JohanMabille I totally agree with you. We should make it right. 

There are multiple places where Matplotlib is a corner case in ipykernel and ipython. 

There are matplotlib tools (displaying logic, formatters setup etc) in ipython: https://github.com/ipython/ipython/blob/a6082c6ff799e4ec30ad24afbfc5f1550185934d/IPython/core/pylabtools.py

There is the `inline` back-end module in ipykernel: https://github.com/ipython/ipykernel/blob/master/ipykernel/pylab

I think this should **not** be in IPython nor ipykernel. It should be in Matplotlib or in a Matplotlib plugin.

AntoineToussaint:
@martinRenou Have you tried using plotly by any chance?

martinRenou:
Just tried, I am coming with a clean fix for that #237

martinRenou:
Replaced by #400

PR #241, opened 2020-04-01
Setting up sys.path dynamically

SylvainCorlay:


.gitignore
```diff
@@ -48,5 +48,8 @@ build/
 # generated kernel specs
 /share/jupyter/kernels/xpython/kernel.json
 
+# generated configuration file
+/share/xeus-python/host
```
SylvainCorlay, line 52: Maybe this should be in `/etc/xeus-python`

src/xsyspath.cpp
```diff
@@ -10,32 +10,51 @@
...
-        Py_SetPath(const_cast<char*>(syspath.c_str()));
-#else
-        static const std::wstring wstr(syspath.cbegin(), syspath.cend());;
-        Py_SetPath(const_cast<wchar_t*>(wstr.c_str()));
-#endif
-#endif
+        std::ifstream host_file(prefix_path() + "/share/xeus-python/host");
+        std::string host_python((std::istreambuf_iterator<char>(host_file) ),
+                                (std::istreambuf_iterator<char>()));
+        host_python.pop_back(); // remove newline
```
SylvainCorlay, line 55: Maybe we could do something more elegant...

I am hesitant to make the `host` file a JSON file, which would prevent this issue, and give room to more keys.

src/xsyspath.cpp
```diff
@@ -8,34 +8,69 @@
...
-            XPYT_STRINGIFY(XEUS_PYTHON_SYSPATH),
-            std::regex("\\{PREFIX\\}"),
-            prefix_path());
+        // Called before Py_Initialize
+        // Python is called with the -S option
+        const std::string syspath = exec(get_host_python() + " -c \"import sys; print(':'.join(sys.path[1:]))\"");
 #if PY_MAJOR_VERSION == 2
-        Py_SetPath(const_cast<char*>(syspath.c_str()));
+        // Py_SetPath is Python 3 only
+        // Py_SetPath(const_cast<char*>(syspath.c_str()));
```
SylvainCorlay, line 70: Py_SetPath is Python 3 only. I guess we won't activate the dynamic syspath flag on Python 2.

SylvainCorlay:
Closing now as superseeded.

PR #254, opened 2020-04-07
Fix ipython display monkey patch and improve the display module

martinRenou:
This patches `IPython.core.display` instead of `IPython.display`. So that we should get `Audio`, `IFrame` etc. for free.

It also implements the missing bits of `IPython.core.display` patch.

### DIFF
--- a/src/xdisplay.cpp
+++ b/src/xdisplay.cpp
@@ -1012,6 +1012,16 @@ namespace xpyt
         }
     }
 
+    py::object pngxy(const py::object& data)
+    {
+        py::module builtins = py::module::import(XPYT_BUILTINS);
+        py::module struct_module = py::module::import("struct");
+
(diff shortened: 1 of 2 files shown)


SylvainCorlay:
How does things behave when IPython is not present? (for what is not missing)

martinRenou:
If IPython is not installed, I suppose

```python
from IPython.core.display import HTML
```

will work, but

```python
from IPython.display import HTML
```

will **not** work.

But I would say that if IPython is not installed, the user should not expect it to work anyway.

Can you actually install Jupyter without ipython?

martinRenou:
@SylvainCorlay That is actually an issue I see with this. ipywidgets and plotly seem to import doing
```python
from IPython.display import display
from IPython.display import display_html
``` 
This means that if we merge this, importing ipywidgets will result in importing IPython as well, which is not the case as of today. What do you think?

martinRenou:
@SylvainCorlay So what do you think? Should we merge this PR?

PR #259, opened 2020-04-14
DisplayObject: Fix reload method

martinRenou:


src/xdisplay.cpp
```diff
@@ -1148,28 +1165,16 @@ namespace xpyt
                 py::arg("data") = py::none(), py::arg("url") = py::none(), py::arg("filename") = py::none(), py::arg("lib") = py::none(), py::arg("css") = py::none())
```
martinRenou, line 1165: This needs to be updated, it needs a custom constructor that calls `reload`

martinRenou:
Closing, as there is no clean way to fix this.

Issue #103, opened 2019-05-28
Display function signature


martinRenou:
The `display` function should have the same signature as the one from IPython. For example the IPython display function takes a list of objects to display, not a single object.

martinRenou:
It is actually not easy to replicate the
```python
def display(*objs, include=None, exclude=None, metadata=None, transient=None, display_id=None, **kwargs):
```
signature, as `pybind11` only allows to have `py::args` at last arguments.

SylvainCorlay:
@martinRenou should we close this?

martinRenou:
Not quite. The signature has been improved but is still not the same.

martinRenou:
This has been fixed

PR #266, opened 2020-04-17
Clean display

martinRenou:
This PR builds against https://github.com/ipython/ipython/pull/12239


martinRenou:
Self-reminder: The plotly support is broken in this PR, due to the `display` function signature which [is still not fixed](https://github.com/jupyter-xeus/xeus-python/issues/103). It should be fixed before merging.

martinRenou:
Replaced by #400

PR #269, opened 2020-05-04
Make output and display redirection optional

lassoan:
Applications may already redirect Python output (for logging, custom console display, etc.). In this cases, the interpreter should not redirect the outputs.
This commit makes output redirection optional (enabled by default).

This pull request implements one potential design, but we are OK with any other option, which allows us to disable output redirection.

### DIFF
--- a/include/xeus-python/xinterpreter.hpp
+++ b/include/xeus-python/xinterpreter.hpp
@@ -38,7 +38,13 @@ namespace xpyt
 
         using gil_scoped_release_ptr = std::unique_ptr<py::gil_scoped_release>;
 
-        interpreter();
+        // If redirect_output_enabled is true (default) then this interpreter will
+        // capture outputs and send them using publish_stream.
+        // Disable this if your interpreter uses custom output redirection.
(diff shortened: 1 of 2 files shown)


include/xeus-python/xinterpreter.hpp
```diff
@@ -66,7 +69,7 @@ namespace xpyt
 
         nl::json internal_request_impl(const nl::json& content) override;
 
-        void redirect_output();
+        void configure_redirect_output();
```
SylvainCorlay, line 72: I am +1 with the feature but I would keep the current method name for backward compatibility - even though it is probably ok given the adoption.

lassoan:
Thanks for the reviews. I reverted to the original method name as suggested.

I also had to add an option to disable display hook installation. We need to keep our display hook for our application's internal Python console widget to work properly, we just call xeus-python's display function from that hook.

SylvainCorlay:
OK, that looks good to me. Merging!

lassoan:
Thank you guys. We are getting very close to switching the xeus-python kernel in Slicer.

PR #270, opened 2020-05-04
Fix binary buffer support

### LINKED ISSUE #268: FileUpload widget
We are switching over to using xeus-python in [3D Slicer](https://www.slicer.org/) and we are very happy with it so far. Thank you very much for developing this library!

I have tried to add a FileUpload iPython widget to a notebook so that we can set up processing workflow that users can apply on their own data. However, FileUpload widget does not return any data. I'm not sure if this is expected or I'm doing something wrong.

How to reproduce:

Run this in a cell:

```python
from ipywidgets import FileUpload
upload = FileUpload()
upload
```

then select a file to upload. After this, `upload.data` and `upload.metadata` fields are set correctly when a regular Python3 kernel is used, but they are empty when xeus-python kernel is used.

martinRenou:
Should fix #268

### DIFF
--- a/src/xutils.cpp
+++ b/src/xutils.cpp
@@ -71,7 +71,7 @@ namespace xpyt
         for (const zmq::message_t& buffer : buffers)
         {
             const char* buf = buffer.data<const char>();
-            bufferlist.attr("append")(py::bytes(buf));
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
Looks good to me! Feel free to go ahead and merge.

PR #277, opened 2020-05-18
backbone for support of magics to xeus-python

btel:
Features to implement:

- [x] magics parsing
- [x] find and execute magics function
- [x] support for line magics
- [x] support for cell magics
- [x] support for shell magics
- [x] support for user defined magics

## Demo

Screen Shot 2020-05-21 at 18 11 17

## Details

the implementation reuses many code parts from IPython:

* `IPython.core.inputtransformers2` - for parsing magics,
* `IPython.core.magic.MagicManager` - for managing magics
* `IPython.core.extensions.ExtensionManager` - for managing extensions
* `IPython.core.magics` - for implementation of magics

The code reimplements the interface of `InteractiveShell` within xeus-python C++ code (see `src/xinteractiveshell.hpp`

In the future, `inputtransformers`, `MagicManager` and `ExtensionManager` could be separated from IPython core so they could be reused for implementation of other kernels. Alternatively, they could be reimplemented in C++ within the xeus or xeus-python code base.

The interface of `InteractiveShell` might be also more clearly defined. Also, the attributes of the shell specific to the magics should be installed by the magic themselves (for easier separation of code, TBD).

### DIFF
--- a/src/xinteractiveshell.cpp
+++ b/src/xinteractiveshell.cpp
@@ -1,0 +1,149 @@
+#include "xinteractiveshell.hpp"
+#include "xdisplay.hpp"
+
+using namespace pybind11::literals;
+namespace py = pybind11;
+
+namespace xpyt
+{
+
+    void xinteractive_shell::init_magics()
+    {
+        m_magic_core = py::module::import("IPython.core.magic");
+        m_magics_module = py::module::import("IPython.core.magics");
+        m_extension_module = py::module::import("IPython.core.extensions");
+
+        m_magics_manager = m_magic_core.attr("MagicsManager")("shell"_a=this);
+        m_extension_manager = m_extension_module.attr("ExtensionManager")("shell"_a=this);
+
+        //shell features required by extension manager
+        m_builtin_trap = py::module::import("contextlib").attr("nullcontext")();
+        m_ipython_dir = "";
+
+        py::object osm_magics =  m_magics_module.attr("OSMagics");
+        py::object basic_magics =  m_magics_module.attr("BasicMagics");
+        py::object user_magics =  m_magics_module.attr("UserMagics");
+        py::object extension_magics =  m_magics_module.attr("ExtensionMagics");
+        m_magics_manager.attr("register")(osm_magics);
+        m_magics_manager.attr("register")(basic_magics);
+        m_magics_manager.attr("register")(user_magics);
+        m_magics_manager.attr("register")(extension_magics);
+        m_magics_manager.attr("user_magics") = user_magics("shell"_a=this);
+
+        //select magics supported by xeus-python
+        auto line_magics = m_magics_manager.attr("magics")["line"];
+        auto cell_magics = m_magics_manager.attr("magics")["cell"];
+        line_magics = py::dict(
+           "cd"_a=line_magics["cd"],
+           "env"_a=line_magics["env"],
+           "set_env"_a=line_magics["set_env"],
+           "pwd"_a=line_magics["pwd"],
+           "magic"_a=line_magics["magic"],
+           "load_ext"_a=line_magics["load_ext"]
+        );
+        cell_magics = py::dict(
+            "writefile"_a=cell_magics["writefile"]);
+
+        m_magics_manager.attr("magics") = py::dict(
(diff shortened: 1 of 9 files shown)


src/xinterpreter.cpp
```diff
@@ -74,10 +79,9 @@ namespace xpyt
         // Monkey patching "from IPython import get_ipython"
         sys.attr("modules")["IPython.core.getipython"] = get_kernel_module();
 
-        // Monkey patching "import linecache". This monkey patch does not work with Python2.
-#if PY_MAJOR_VERSION >= 3
-        sys.attr("modules")["linecache"] = get_linecache_module();
-#endif
+
+        // add get_ipython to global namespace
+        exec(py::str("from IPython.core.getipython import get_ipython"));
```
martinRenou, line 84: You could also do:

```python
py::globals()["get_ipython"] = get_kernel_module();
```

I fear that the `exec` could end up importing IPython if not executed after the above line.

src/xinterpreter.cpp
```diff
@@ -116,6 +120,10 @@ namespace xpyt
             return kernel_res;
         }
 
+        py::module input_transformers = py::module::import("IPython.core.inputtransformer2");
+        py::object transformer_manager = input_transformers.attr("TransformerManager")();
+        py::str code_copy = transformer_manager.attr("transform_cell")(code);
```
martinRenou, line 125: How hard do you think it would be to re-implement this in xeus-python?

Would it make more sense to implement it in C++ or should we extract the inputtransformer in a separate package to get rid of the IPython dependency?

btel: we need to investigate, but the parsing logic is rather nontrivial. For version 0, I would simply fail silently if IPython is not installed (and so the magic would not be supported without IPython installed).

btel: separating into a separate package is also an option, but I would not do that without discussing with IPython maintainers first (to avoid maintaining an independent code).

martinRenou: > I would simply fail silently if IPython is not installed

Does it kill the kernel? It might be good if we show a warning saying that IPython must be installed in order to use magics.

src/xinterpreter.cpp
btel: good point, but in this case we need to be careful. `get_kernel_module` should be called only once, because we should only create one instance of InteractiveShell. We might fix that by implementing `InteractiveShell` as a singleton, but it's not yet done.

src/xinterpreter.cpp
btel: my idea is to try importing and if it fails, we will not parse the magics (and therefore no magic support). Of course, we might print a warning to the stdout or elsewhere.

src/xinterpreter.cpp
martinRenou: Actually it can be called as many times as you want, it will create the `InteractiveShell` only once. This is ensured by the `static` at this line: https://github.com/jupyter-xeus/xeus-python/blob/ec7338cafbe0f172bf9f1dcd0c0088713b35d167/src/xcomm.cpp#L226 that will create the module instance only at the first time the function is called. Next times it will return the already created module.

src/xinteractiveshell.cpp
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    void xinteractive_shell::register_magics(py::args args)
+    {
+        m_magics_manager.attr("register")(*args);
+    }
+
+    // define getters
+
+    py::object xinteractive_shell::get_magics_manager()
```
JohanMabille, line 114: even if python cast aways const-correctness, I think it could be usefull to mark this method as `const` since it is bound as a redaonly property. The same applies ot methods below.

src/xinterpreter.cpp
btel: oh, nice. I didn't know you could use `static` with variables.

martinRenou: Yep :) It's really nice: https://www.studytonight.com/cpp/static-keyword.php

src/xinteractiveshell.cpp
btel: thanks! done

PR #284, opened 2020-06-15
Moved pybind11 headers into a single header and undef Python snprintf…

### LINKED ISSUE #283: Wrap pybind11 headers
There is a [bug](https://bugs.python.org/issue36020) in Python 3.7 and 3.8 implementation where `snprintf` is defined as a macro on Windows. This definition is pulled in the code by pybind11 headers. This leads to compilation errors on recent Windows compilers when including other libraries that use `std::snprintf`.

A solution is to wrap pybind11 headers in a single header that undefines the macro. This should be done after #279 is merged.

JohanMabille:
… macro

Fixes #283


JohanMabille:
The problem has been fixed in pybind11, so let's close this (otherwise we would need to duplicate the fix in pybind11_json) and wait for a release of pybind11.

PR #288, opened 2020-06-19
Optional ipython

JohanMabille:
@martinRenou I think the kernel_module initialization is a bit cleaner with this PR.

I would have prefer to achieve the memoization on the Python side, (see the first commit), however this leads to random crash when executing Python code (unrelated to widgets) in the notebook.

### DIFF
--- a/src/xcomm.cpp
+++ b/src/xcomm.cpp
@@ -65,6 +65,13 @@ namespace xpyt
         xeus::xcomm m_comm;
     };
 
+    struct xcomm_manager
+    {
+        xcomm_manager() = default;
+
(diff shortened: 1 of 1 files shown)


src/xcomm.cpp
```diff
@@ -356,25 +344,86 @@ namespace xpyt
...
+    {
+        ipython_instance() : m_instance(py::none())
+        {
+        }
+
+        py::object get_instance(const py::module& kernel_module) const
+        {
+            if (m_instance.is(py::none()))
+            {
+                // The first import of IPython will throw is IPython has not been installed.
```
martinRenou, line 377: will throw "if" IPython is not installed

PR #289, opened 2020-06-22
Loads IPython only if available

JohanMabille:
We still need to fix IPyWidgets.

### DIFF
--- a/src/xinterpreter.cpp
+++ b/src/xinterpreter.cpp
@@ -92,6 +92,12 @@ namespace xpyt
 
         // Add get_ipython to global namespace
         py::globals()["get_ipython"] = get_kernel_module().attr("get_ipython");
+        
+        // Initializes get_ipython result
+        get_kernel_module().attr("get_ipython")();
+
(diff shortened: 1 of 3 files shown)


src/xinterpreter.cpp
```diff
@@ -92,6 +92,9 @@ namespace xpyt
 
         // Add get_ipython to global namespace
         py::globals()["get_ipython"] = get_kernel_module().attr("get_ipython");
+
+        m_has_ipython = get_kernel_module().attr("has_ipython").cast<bool>();
```
martinRenou, line 96: Where do you set the `has_ipython` value in the kernel module?

JohanMabille: [Here](https://github.com/jupyter-xeus/xeus-python/pull/289/files#diff-d1e8fed419dfbdd955a412b7b0111e28R389) or [here](https://github.com/jupyter-xeus/xeus-python/pull/289/files#diff-d1e8fed419dfbdd955a412b7b0111e28R384)

martinRenou: Oh yeah sorry I missed it

PR #290, opened 2020-06-26
Advertize python version in kernelspec

SylvainCorlay:
Instead of showing `xpython`, we will now show e.g. `Python 3.7 (XPython)`.


CMakeLists.txt
```diff
@@ -64,7 +69,7 @@ OPTION(XPYT_DOWNLOAD_GTEST "build gtest from downloaded sources" OFF)
 # Dependencies
 # ============
 
-set(xeus_REQUIRED_VERSION 0.23.12)
+set(xeus_REQUIRED_VERSION 0.24.0)
```
SylvainCorlay, line 72: This was outstanding from a previous PR. We do require 0.24 now because of the history API changes.

SylvainCorlay:
@afshin FYI

PR #307, opened 2020-07-15
Update README

### LINKED ISSUE #306: Update readme to include magics support
The readme says that IPython magics are not supported whichbis not the case anymore.

martinRenou:
Should fix #306

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,16 +109,17 @@ link:
 
 To get started with using `xeus-python`, check out the full documentation
 
-http://xeus-python.readthedocs.io/
+http://xeus-python.readthedocs.io
 
 ## What are the advantages of using xeus-python over ipykernel (IPython kernel)?
 
 Check-out this blog post for the answer: https://blog.jupyter.org/a-new-python-kernel-for-jupyter-fcdf211e30a8.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -109,16 +109,17 @@ link:
...
 
-http://xeus-python.readthedocs.io/
+http://xeus-python.readthedocs.io
 
 ## What are the advantages of using xeus-python over ipykernel (IPython kernel)?
 
 Check-out this blog post for the answer: https://blog.jupyter.org/a-new-python-kernel-for-jupyter-fcdf211e30a8.
 Long story short:
-xeus-python does not cover 100% of the features of ipykernel. For examples, IPython magics are not supported yet by xeus-python. However:
+xeus-python does not cover 100% of the features of ipykernel. For examples, Matplotlib is not supported yet by xeus-python and only some magics are supported for now. However:
```
SylvainCorlay, line 118: We should probably say that the matplotlib magic is not supported (and nothing prevents using matplotllib as a library).

Also "for example" should probably not have an s.

martinRenou:
@SylvainCorlay I took your comment into account

PR #310, opened 2020-07-21
Use Py_SetProgramName

SylvainCorlay:
Another way to do things would be to use `PySys_SetArgvEx`, but we still need to convert argv to `wchar_t`.

Another way to do things instead of converting would be to use `wmain` instead of `main`. I am not sure how standard that is.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -101,6 +101,11 @@ int main(int argc, char* argv[])
     signal(SIGSEGV, handler);
 #endif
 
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
Closing for now.

JohanMabille:
I think the `PySys_SetArgvEx` is more flexible (we may need additional arguments from the command line on the Python side later).

From what I know, `wmain` is specific to Windows.

PR #319, opened 2020-09-04
Enable multiple display

SylvainCorlay:


SylvainCorlay:
This may require some cleanup.

src/xinteractiveshell.hpp
```diff
@@ -20,7 +20,8 @@ namespace xpyt
     {
         static inline void show_in_pager(py::str data, py::kwargs)
         {
-            xdisplay(py::dict("text/plain"_a=data), {}, {}, py::dict(), py::none(), py::none(), false, true);
+            py::kwargs kw = py::dict(py::arg("raw") = true);
+            xdisplay(py::dict("text/plain"_a=data), kw);
```
SylvainCorlay, line 24: Need to check that I can pass a single dict to a function taking py::args.

src/xdisplay.cpp
```diff
@@ -188,120 +188,160 @@ namespace xpyt
...
+                {
+                    interp.display_data(pub_data, pub_metadata, std::move(cpp_transient));
+                }
             }
         }
     }
 
+    void xdisplay(py::args objs, py::kwargs kw)
+    {
+        // Same as xdisplay - passing update = true to xdisplay_impl
```
martinRenou, line 248: You might want to remove this comment or fix it to `// Same as xupdate_display - passing update = false to xdisplay_impl`

SylvainCorlay: Done.

PR #326, opened 2020-10-03
Upgraded to xeus 0.24.2

JohanMabille:


SylvainCorlay:
It weird that build 0 of cppzmq is being pulled

JohanMabille:
We need to rebuild pyzmq because it vendors zeromq on Windows.

SylvainCorlay:
It does not. see my message and my pyzmq feedstock pr.

JohanMabille:
Looks like something went wrong when rebuilding pyzmq-feedstock, it used zeromq 4.3.2 instead of zeromq 4.3.3.

PR #328, opened 2020-10-08
Bugfix: check for string inside include vector

### LINKED ISSUE #324: Invoking `IPython.display.display_html` on an HTML object renders repr representation
The following example
```python
import IPython.display
IPython.display.display_html(   
    IPython.display.HTML("<i>an html element</i>")
)
```
produces  `<display.HTML object at 0x7f33a5450630>`

agoose77:
(Hopefully) fixes #324


martinRenou:
LGTM! Thanks a lot

PR #331, opened 2020-10-23
Add support for extensions

martinRenou:
This implements support for IPython extensions. They must be installed under `$PREFIX/etc/xeus-python/extensions/` and have the following JSON format:
```json
{
    "module": "module_name",
    "enabled": true
}
```


martinRenou:
As discussed in-person, we should not rely on `xpath` in `xinterpreter`. Also let's make it a `load_extensions` method.

martinRenou:
@JohanMabille I took your review into account

JohanMabille:
Hey, I totally forgot this, sorry for the late merge!

PR #332, opened 2020-10-23
Add user_global_ns property to the Interactive shell

martinRenou:
This seems to be used by some custom magics


martinRenou:
Let's merge this, the CI issue is not caused by this PR.

PR #341, opened 2020-10-28
Clean xpath prefix

martinRenou:
The prefix path was resulting in `$PREFIX/bin/../`, this PR changes the prefix path to `$PREFIX/`


martinRenou:
@SylvainCorlay do you agree with this change?

PR #343, opened 2020-10-30
Fix XPYT_DISABLE_TUNE_GENERIC

chronitis:
There appeared to be some logic missing in this section which meant that if you set both `XPYT_DISABLE_ARCH_NATIVE` and `XPYT_DISABLE_TUNE_GENERIC`, you still got `-mtune=generic` in `CPPFLAGS` and consequent build failures on architectures for which this flag doesn't work.

This reuses the logic found elsewhere in this file which appeared to correctly handle these two flags.


JohanMabille:
Good catch! Thanks for the fix.

PR #355, opened 2020-11-16
Pass sys.executable in debugger config

SylvainCorlay:
Depends on https://github.com/jupyter-xeus/xeus/pull/259.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -102,14 +102,11 @@ int main(int argc, char* argv[])
     static const std::string executable(XEUS_PYTHON_EXECUTABLE);
     static const std::wstring wexecutable(executable.cbegin(), executable.cend());
 
-// On windows, sys.executable is not properly set with Py_SetProgramName
-// Cf. https://bugs.python.org/issue34725
-// A private undocumented API was added as a workaround in Python 3.7.2.
-#ifdef _WIN32
-    _Py_SetProgramFullPath(const_cast<wchar_t*>(wexecutable.c_str()));
-#else
+    // On windows, sys.executable is not properly set with Py_SetProgramName
+    // Cf. https://bugs.python.org/issue34725
+    // A private undocumented API was added as a workaround in Python 3.7.2.
+    // _Py_SetProgramFullPath(const_cast<wchar_t*>(wexecutable.c_str()));
     Py_SetProgramName(const_cast<wchar_t*>(wexecutable.c_str()));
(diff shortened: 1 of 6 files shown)


SylvainCorlay:
@JohanMabille do you think we should merge as-is?

src/xdebugger.cpp
```diff
@@ -128,8 +130,16 @@ namespace xpyt
...
     {
-        // debugpy has to be started in the main thread
-        std::string code = "import debugpy\ndebugpy.listen((\'" + m_debugpy_host + "\'," + m_debugpy_port + "))";
+        // import debugpy 
+        std::string code = "import debugpy;";
+        // specify sys.executable
+        auto it = m_debugger_config.find("python");
+        if (it != m_debugger_config.end())
+        {
+            code += "debugpy.configure({\'python\': \'" + it->template get<std::string>()  + "\'});";
```
JohanMabille, line 139: I think you can remove the `;` before the last `"`

src/main.cpp
```diff
@@ -102,14 +102,11 @@ int main(int argc, char* argv[])
...
-// On windows, sys.executable is not properly set with Py_SetProgramName
-// Cf. https://bugs.python.org/issue34725
-// A private undocumented API was added as a workaround in Python 3.7.2.
-#ifdef _WIN32
-    _Py_SetProgramFullPath(const_cast<wchar_t*>(wexecutable.c_str()));
-#else
+    // On windows, sys.executable is not properly set with Py_SetProgramName
+    // Cf. https://bugs.python.org/issue34725
+    // A private undocumented API was added as a workaround in Python 3.7.2.
+    // _Py_SetProgramFullPath(const_cast<wchar_t*>(wexecutable.c_str()));
```
JohanMabille, line 108: Should we call this private API when Python's version is greater than 3.7.1?

SylvainCorlay: We could do it separately.

src/xdebugger.cpp
SylvainCorlay: Don't we do `code += "debugpy.listen` right after?

JohanMabille: Ha my bad, I was persuaded the statements were separated with ' \n'.

src/main.cpp
JohanMabille: Ok!

PR #367, opened 2020-12-07
Put xutils and xtraceback in the public API

martinRenou:


JohanMabille:
You should add Windows export / import macro to the declaration of classes and functions of the public API.

martinRenou:
Ok, thanks for the review.

We should maybe not put everything in the public API. Only ~4 functions are useful for downstream projects.

JohanMabille:
Fox xutils.hpp, indeed. I let you add what you need for xeus-robot, I will make a pass for the debugger.

JohanMabille:
LGTM

PR #368, opened 2020-12-08
Fix pybind11 deprecation warnings 

### LINKED ISSUE #365: Deprecation warnings from pybind11
Building xeus-python locally, I am getting a few warnings concerning deprecation of the `py::module` constructor:
```
/home/martin/github/xeus-python/src/xlinecache.cpp:56:48: warning: 'pybind11::module_::module_(const char*, const char*)' is deprecated: Use PYBIND11_MODULE or module_::create_extension_module instead [-Wdeprecated-declarations]
   56 |         py::module linecache_module("linecache");
      |                                                ^
In file included from /home/martin/github/xeus-python/src/xlinecache.cpp:13:
/home/martin/miniconda3/envs/xeus-robot/include/pybind11/pybind11.h:897:14: note: declared here
  897 |     explicit module_(const char *name, const char *doc = nullptr) {
      |              ^~~~~~~
```

martinRenou:
Fix #365


SylvainCorlay:
Hum, too bad it is deprecated. I found the previous API simpler.

How about we keep the previous API call in a comment, and ask upstream if they would be OK de-deprecating it.

martinRenou:
I agree, I opened a PR simplifying it a bit: https://github.com/pybind/pybind11/pull/2717

SylvainCorlay:
Ah, I was opening https://github.com/pybind/pybind11/issues/2718 at the same time.

src/xinternal_utils.cpp
```diff
@@ -32,6 +32,11 @@ namespace nl = nlohmann;
 
 namespace xpyt
 {
+    py::module create_module(const std::string& module_name)
+    {
+        return py::module_::create_extension_module(module_name, nullptr, new py::module_::module_def);
```
JohanMabille, line 37: should be 

```cpp
return py::module_::create_extension_module(module_name.c_str(), nullptr, new py::module_::module_def);
```

PR #372, opened 2020-12-10
Fixed relocation issue of executable name

JohanMabille:


src/xpaths.cpp
```diff
@@ -34,6 +33,12 @@
 #  include <stdlib.h>
```
martinRenou, line 33: Could we make this file part of the library? So the code is not duplicated in xeus-robot? Otherwise we'll need to port those changes to xeus-robot as well

JohanMabille: I'm actually gonne split ths file, a part is going to be part of the XTL, the other part would be moved into the xeus-python  library.

SylvainCorlay:
We should use the xtl version in xeus-cling as well now.

PR #376, opened 2020-12-14
Handle empty string for pythonhome

martinRenou:


martinRenou:
Arf, let's remove pytest here as well, I don't trust it

SylvainCorlay:
We need a better kernel testing tool *à la* `jupyter_kernel_test`.

PR #379, opened 2021-01-05
remove the promise to implement debugger

akhmerov:
A sentence in readme wasn't updated from before the debugger was released.


JohanMabille:
Thanks!

PR #390, opened 2021-01-15
Require ipywidgets>=7.6 for JupyterLab 3 support

jtpio:
So the widgets work out of the box on Binder with JupyterLab 3.0.


martinRenou:
Thanks!

PR #395, opened 2021-01-27
Add --version option to the command line interface

martinRenou:


src/main.cpp
```diff
@@ -49,6 +50,18 @@ void handler(int sig)
 
 namespace py = pybind11;
 
+bool should_print_version(int& argc, char* argv[])
```
JohanMabille, line 53: `argc` should be passed by vlaue

PR #400, opened 2021-02-01
Use IPython

martinRenou:
Fix #62, fix #63, partially fix #224, fix #28, fix #359, fix #399, fix #402

PR #401, opened 2021-02-01
Matplotlib magics

martinRenou:
Just for heads up, it seems to not work for some reason, I'll continue iterating on this


martinRenou:
Replaced by #400

PR #414, opened 2021-02-17
Readd jupyter-kernel-test

SylvainCorlay:


test/test_xeus_python_kernel.py
```diff
@@ -0,0 +1,130 @@
...
+        {'text': 'se', 'matches': {'set', 'setattr'}},
+    ]
+
+    complete_code_samples = ['1', "print('hello, world')", "def f(x):\n  return x*2\n\n\n"]
+    incomplete_code_samples = ["print('''hello", "def f(x):\n  x*2"]
+    invalid_code_samples = ['import = 7q']
+
+    code_inspect_sample = "open"
+
+    def test_xeus_python_line_magic(self):
```
martinRenou, line 37: We might want to remove all magics tests. We don't want our tests to break if IPython changes some magics logs etc.

PR #425, opened 2021-03-09
Change cmake build in README to out-of-source

telamonian:
This is kind of a dumb thing to submit a PR for, but it caught my eye. The current `cmake` instructions in the README result in an in-source build. In-source builds are considered to be bad practice, and bad advice to give beginners. I changed the instructions to instead create an out-of-source build.

There are a number of reasons to prefer out-of-source over in-source, but the big one is that the resulting `build` dir is easy to nuke and recreate, no matter how badly you screw things up.


JohanMabille:
Good catch! Totally agree with you, this has been our workflow since the beginning, I don't know why we didn't reflect this in the install instructions. Anyway, thanks for fixing it!

PR #446, opened 2021-05-26
Implemented rich inspect variables request

JohanMabille:
@fcollonval the content of the debug request to send is:

```
{
    "type": "request", 
    "seq":  int, 
    "command": "richInspectVariables",
    "body": {
        "onBreakpoint", bool,
        "variableName": string,
        "variableReference", int
    }
}
```
The "onBreakpoint" field is temporary, I need to release xeus so I can get it out. For now, this only works when the debugger did not stop on a breakpoint. In this case, you can omit the `variableReference` field.

### DIFF
--- a/include/xeus-python/xdebugger.hpp
+++ b/include/xeus-python/xdebugger.hpp
@@ -29,6 +29,8 @@ namespace xpyt
     {
     public:
 
+        using base_type = xeus::xdebugger_base;
+
         debugger(zmq::context_t& context,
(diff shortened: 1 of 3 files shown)


fcollonval:
Thanks @JohanMabille
I'll update the JLab PR: https://github.com/jupyterlab/jupyterlab/pull/10299

JohanMabille:
@fcollonval you can now remove the `onBreakpoint` field in the request.

PR #447, opened 2021-05-26
Fix Output widget

### LINKED ISSUE #445: Output widget not working
output

martinRenou:
Fix #445


martinRenou:
This is not really correct. We should make it a proper `XKernel` property that gets computed dynamically, otherwise it does not work in other contexts than the execute request.

PR #453, opened 2021-06-03
Added test for setExceptionBreakpoint request

JohanMabille:


test/test_debugger.cpp
```diff
@@ -328,6 +328,27 @@ nl::json make_rich_inspect_variables_request(int seq, const std::string& var_nam
...
+    nl::json req = {
+        {"type", "request"},
+        {"seq", seq},
+        {"command", "setExceptionBreakpoints"},
+        {"arguments", {
+            {"filters", {"raised", "uncaught"}},
+            {"exceptionOptions", options}
+        }}
+    };
+    return req;
```
JohanMabille, line 349: @goanpeca This is the request you need to send to have debugpy break on exceptions.

goanpeca: Thanks @JohanMabille !

PR #467, opened 2021-08-19
Refactor xeus-python-shell

martinRenou:
Using https://github.com/martinRenou/xeus-python-shell

This removes all the ugly `py::exec` calls


JohanMabille:
Nice refactoring. However, I think the python files from `xeus-python-shell` should be part of this repo.

JohanMabille:
I changed my mind given the issues we would have trying to install python files from a C++ package. So this should be merged once the issue on Windows is fixed.

SylvainCorlay:
@meeseeksdev backport to 0.12.x

PR #471, opened 2021-09-01
Pin right IPython version on appveyor

martinRenou:
This fixes the environment solving issue on appveyor


martinRenou:
Continuing this in https://github.com/jupyter-xeus/xeus-python/pull/472, there are multiple issues on the CI now, let's fix them all in one PR

PR #472, opened 2021-09-01
Fix CI

martinRenou:
- We shouldn't test the content of the error message, as it's coming from IPython, it's not on us to test it.
- Fix jupyter_client version pulled by jupyter_kernel_test (should be fixed upstream)
- Fix IPython version used in the CI


martinRenou:
Apart from the random failure that causes the test to never finish, it seems to be good to go @JohanMabille

PR #474, opened 2021-09-03
Handle shutdown from jupyter_client

martinRenou:


src/main.cpp
```diff
@@ -48,6 +48,11 @@ void handler(int sig)
 }
 #endif
 
+void stop_handler(int /*sig*/)
+{
+    exit(0);
```
martinRenou, line 53: @JohanMabille @SylvainCorlay should we do a proper `server.stop` in the handler?

JohanMabille: I think it's not possible, you are very limited in what you can do inside a signal handler.

martinRenou: Yeah that's what I was seing... 

I see that the xserver in xeus does not implement a destructor, should we implement a destructor that calls `stop`?

JohanMabille: No, you mustn't call a virtual method from a destructor (`stop calls `stop_impl` in the end). And the `stop_impl` method may throw, which would prevent the destruction of the sockets.

martinRenou: Ok. So I guess my question now is: is it fine to `exit(0)` without calling `server.stop()`?

JohanMabille: Yes.

PR #484, opened 2021-09-13
Upgrade to xeus 2.O.0

JohanMabille:


.appveyor.yml
```diff
@@ -27,7 +27,7 @@ install:
   - conda update -q conda
   - conda info -a
   # Host dependencies
-  - conda install xeus=1.0.3 nlohmann_json cppzmq xtl "pybind11>=2.6.1,<3.0" "pybind11_json>=0.2.6,<0.3" gtest=1.8.0 debugpy "xeus-python-shell>=0.1.5,<2" "jupyter_client<=6.2" -c conda-forge
+  - conda install xeus=2.0.0 nlohmann_json cppzmq xtl pybind11=2.6.0 pybind11_json=0.2.8 gtest=1.8.0 debugpy "xeus-python-shell>=0.1.5,<2" "jupyter_client<=6.2" -c conda-forge
```
martinRenou, line 30: Why not keeping `"pybind11>=2.6.1,<3.0"` and others? I changed this for having the same test environment as on other platforms.

Ideally we would even use the `dev-environment.yml` file

martinRenou: Also those are the pins as on the feedstock. It's safer to pin the same versions here so that we realize early if something goes wrong.

PR #492, opened 2021-09-22
Add raw mode

trungleduc:
## New features

This PR adds a "raw" execution mode to `xeus-python`, in this mode, `xeus-python` executes code directly without using `ipython`. This is how `xeus-python` works before version `0.11.0` .

By using this mode, users who do not need `ipython` magics can accelerate the startup time of kernel. 

To start `xeus-python` kernel in raw mode, just add `-r` or `--raw` option to startup arguments, for example:
```json
{
  "display_name": "Python 3.9 (XPython Raw)",
  "argv": [
      ".../bin/xpython",
      "-f",
      "{connection_file}",
      "--raw"
  ],
  "language": "python",
  "metadata": { "debugger": false }
}
```

## Benchmark

- Voila started with:
```
voila test.ipynb --no-browser
```
`test.ipynb` is a notebook with just one `print('')` command, 3 kernels is used: `ipython`, `xeus-python` and `xeus-python (raw mode)`

- Benchmark scenario: perform 50 requests to `Voila`, 10 concurrent requests at a time .  
```
 ab -n 50 -c 10 http://localhost:8868/
```
- Result:
  -  `ipython` :  Requests per second:    4.93 [#/sec], Time per request:  2028.115 [ms] 
  -  `xeus-python` :  Requests per second:     5.11 [#/sec], Time per request:  1958.001 [ms] 
  -  `xeus-python (raw mode)` :  Requests per second:   7.98 [#/sec], Time per request:  1253.689 [ms] 

##  Current status

- [x] CLI parameter to switch between normal and raw mode.
- [x] Code execution. 
- [x] Code completion.
- [x] Code inspection.
- [x] Patch `IPython.core.display`

## Backwards-incompatible changes

N/A


src/xinterpreter_raw.cpp
```diff
@@ -0,0 +1,348 @@
...
+                {"data", {
+                    {"text/plain", result}
+                }},
+                {"source", "page"},
+                {"start", 0}
+            });
+            kernel_res["user_expressions"] = nl::json::object();
+
+            return kernel_res;
+        }
```
martinRenou, line 122: I would not keep this fake support, let's keep it simple and just do the execution

src/xinterpreter_raw.cpp
```diff
@@ -0,0 +1,348 @@
...
+        py::object transformer_manager;
+        py::list result = transformer_manager.attr("check_complete")(code);
+        auto status = result[0].cast<std::string>();
+
+        kernel_res["status"] = status;
+        if (status.compare("incomplete") == 0)
+        {
+            kernel_res["indent"] = std::string(result[1].cast<std::size_t>(), ' ');
+        }
+        return kernel_res;
```
martinRenou, line 251: I believe this doesn't work.

Let's not implement this request for now. It will not be used by Voila or JupyterLab anyway, it's only for the console.

martinRenou (commented):
Looking pretty good 😃

src/xinterpreter_raw.cpp
```diff
@@ -324,4 +322,26 @@ namespace xpyt
...
     }
 
+
+    std::string raw_interpreter::remove_magics(const std::string& code, bool warning)
+    {
+        std::string result = "";
+        std::stringstream stream(code);
+        std::string temp;
+        while (std::getline(stream, temp)) {
+            if (temp.rfind("%", 0) != 0) {
```
martinRenou, line 332: I believe this doesn't work. For example the following code cell will not work though it's valid Python:

```python
a = ("hey %x"
% 3)
```

This fails with a SyntaxError because the cleaned code ends up being:
```python
a = ("hey %x"
import logging
logging.getLogger().warning('IPython magics are disabled in raw mode')
```
which is not valid

martinRenou: I guess you want to check only the first non-empty line of code for magics?

Is it valid to have a magic in the middle of the code anyway?

martinRenou (changes requested):

src/xinterpreter_raw.cpp
trungleduc: Indeed my logic does not work well, it'll fix it, thanks @martinRenou

martinRenou: Mmh that's annoying. I wonder if we could strip down the IPython's input transformer and reuse it a bit?

martinRenou: Or it might be better to just let the execution fail with a `SyntaxError`

trungleduc:
```
---------------------------------------------------------------------------
SyntaxError                               Traceback (most recent call last)
File /home/trungle/miniconda3/envs/xpydev/lib/python3.9/ast.py, in parse:
Line 50:    return compile(source, filename, mode, flags,

SyntaxError: invalid syntax (<string>, line 2)
---------------------------------------------------------------------------
There may be Ipython magics in your code, this feature is not supported in xeus-python raw mode!
```

@martinRenou  something like this will do the job ? it will add a  warning on `SyntaxError` with `%` symbol in code.

martinRenou: 👍🏽 Yep

We can also suggest the user to either use classic xeus-python or get rid of the magic

src/xkernel.cpp
```diff
@@ -74,14 +75,190 @@ namespace xpyt
...
+        xpyt::xcomm_manager m_comm_manager;
+    };
+
+    /*****************
+     * kernel module *
+     *****************/
+
+    void bind_history_manager(py::module& kernel_module)
+    {
+        py::class_<xeus::xhistory_manager>(kernel_module, "HistoryManager")
```
martinRenou, line 111: I am not sure we actually need this, IIRC this was for the custom magics we had before depending on IPython. But let's keep it for now and see later if we can remove it.

trungleduc:
@martinRenou  the brackets positions are updated.

martinRenou (approved):

PR #508, opened 2021-10-19
Early returns of richInspectVariable for invalid variable names

### LINKED ISSUE #475: Should the debugger check validity of expressions passed to `richInspectVariables`
Currently wee can get a syntax error like:

```
Error processing internal command.
Traceback (most recent call last):
  File "/home/krassowski/miniconda3/envs/xeus-python/lib/python3.9/site-packages/debugpy/_vendored/pydevd/_pydevd_bundle/pydevd_vars.py", line 419, in evaluate_expression
    compiled = compile(_expression_to_evaluate(expression), '<string>', 'eval')
  File "<string>", line 1
    function variables_repr_data,function variables_repr_metada = (get_ipython().display_formatter.format(function variables))
             ^
SyntaxError: invalid syntax
```

if client does a silly thing of passing not-a-variable to the kernel debugger (see https://github.com/jupyterlab/jupyterlab/issues/11009). While the error is on the client side in the first place, should the debugger reject this nicely rather than error out?

JohanMabille:
Fixes #475

### DIFF
--- a/src/xdebugger.cpp
+++ b/src/xdebugger.cpp
@@ -72,48 +72,6 @@ namespace xpyt
         p_debugpy_client = nullptr;
     }
 
-    namespace
-    {
-        using name_list = std::vector<std::string>;
-
-        const name_list& get_excluded_variables()
-        {
-            static name_list l =
-            { 
-                "__name__",
-                "__doc__",
-                "__package__",
-                "__loader__",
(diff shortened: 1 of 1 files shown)


JohanMabille:
The issue is due to jupyter_kernel_tests

PR #523, opened 2022-02-26
cmake: Support compiling with an absolute PYTHONHOME

doronbehar:
Useful on systems such as NixOS and Guix where Python is installed in a
totally different directory which can't refer to with a relative path.


JohanMabille:
Thanks!

PR #530, opened 2022-04-28
Add WASM build

DerThorsten:
- work in progress for jupyterlite / emscripten-forge dev.


src/main_emscripten_kernel.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+#include <emscripten/bind.h>
+
+#include "xeus-python/xinterpreter.hpp"
+#include "xeus/xembind.hpp"
+
+
+EMSCRIPTEN_BINDINGS(my_module) {
+    xeus::export_core();
+    using interpreter_type = xpyt::interpreter;
```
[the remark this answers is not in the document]
DerThorsten: because it runs out of scope.
The EMSCRIPTEN_BINDINGS scope is shortlifed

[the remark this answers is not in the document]
JohanMabille: The `scoped_interpreter` object is movable. Another solution would be to have a class that inherits from the interpreter class and implements the features specific to WASM. I really don't like propagating this conditional compilation everywhere.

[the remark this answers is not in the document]
JohanMabille: Actually it is the same for the CMakeLists, we should add targets specifi to WASM instead of conditions everywhere.

martinRenou:
I am iterating on the CMakeLists to make a separate target for xeus-python-wasm

martinRenou:
@DerThorsten @JohanMabille I pushed some commits.

- Add an interpreter_wasm class
- Add a custom target for xeus-python-wasm (lib) and xpython_wasm (executable)
- Added a CI job for compiling to wasm (cannot pass yet, we need to wait for the emscripten-forge)

build_all.sh
```diff
@@ -0,0 +1,39 @@
+#bin/bash
```
martinRenou, line 1: Adding a note that we need to remove this file before merging the PR

martinRenou:
@DerThorsten Github doesn't let me request a review from you, so I'm pinging you here :)

martinRenou:
Rebased and now the CI passes 🚀

overwriteProp.cmake
```diff
@@ -0,0 +1,6 @@
+# https://github.com/emscripten-core/emscripten/issues/15276
```
JohanMabille, line 1: This flie should be in a cmake folder, not a the root of the project.

include/xeus-python/xinterpreter.hpp
```diff
@@ -73,6 +73,7 @@ namespace xpyt
         nl::json internal_request_impl(const nl::json& content) override;
 
         void redirect_output();
+        void instanciate_ipython_shell();
```
JohanMabille, line 76: Since you redefine it in the `wasm_interpreter` class, this method should be virtual (and use override in the inheriting class to avoid signature silent errors). Besides, if the implementations of this method in inheriting classes never calls the implementation of the mother class, this method should probably be private.

CMakeLists.txt
```diff
@@ -315,6 +380,66 @@ if (XPYT_BUILD_XPYTHON_EXTENSION)
...
+
+    target_compile_options(xpython_wasm
+        PUBLIC --std=c++17
+        PUBLIC -Wno-deprecated
+        PUBLIC "SHELL: --bind"
+        PUBLIC "SHELL: -s ENVIRONMENT=web"
+        PUBLIC "SHELL: -fwasm-exceptions"
+        PUBLIC "SHELL: -s FORCE_FILESYSTEM"
+        PUBLIC "SHELL:  -s LZ4=1"
+        # PUBLIC "SHELL: -s EXPORTED_FUNCTIONS='[\"_post_kernel_message\"]'"
```
JohanMabille, line 415: I think a check to remove unecessary flags is needed here.

CMakeLists.txt
```diff
@@ -315,6 +380,66 @@ if (XPYT_BUILD_XPYTHON_EXTENSION)
...
+        PUBLIC "SHELL: -fwasm-exceptions"
+        # PUBLIC "SHELL: -s ENVIRONMENT=${XEUS_EMSCRIPTEN_ENVIRONMENT}"
+        PUBLIC "SHELL: -s MAIN_MODULE=1"
+        PUBLIC "SHELL: -s ENVIRONMENT=worker"
+        PUBLIC "SHELL: -s TOTAL_STACK=512mb"
+        PUBLIC "SHELL: -s INITIAL_MEMORY=1024mb"
+        PUBLIC "SHELL: -s FORCE_FILESYSTEM"
+        PUBLIC "SHELL:  -s LZ4=1"
+        PUBLIC "SHELL: --post-js post.js"
+        # PUBLIC "SHELL: -s EXPORTED_FUNCTIONS='[\"_post_kernel_message\"]'"
```
JohanMabille, line 441: Same remark here.

patch_it.py
```diff
@@ -0,0 +1,9 @@
+with open('./build/xpython_wasm.js', 'r') as f:
+    content = f.read()
+
+
+query = 'Module["preloadPlugins"].push(audioPlugin);'
+content = content.replace(query, '')
+
+with open('./build/xpython_wasm.js', 'w') as f:
+    f.write(content)
```
JohanMabille, line 9: Why do we need this?

post.js
```diff
@@ -0,0 +1,3 @@
+if (!('wasmTable' in Module)) {
+    Module['wasmTable'] = wasmTable
+}
```
JohanMabille, line 3: Why do we need this?

JohanMabille (changes requested):
Some points are still unclear to me, let's discuss this tomorrow;

martinRenou:
Remaining points are:
- Resolve the comments concerning the flags
- Wait for https://github.com/conda-forge/xeus-python-shell-feedstock/pull/11 to be published so that the CI can pass

PR #538, opened 2022-05-30
added pyjs

DerThorsten:


src/xinterpreter_wasm.cpp
```diff
@@ -24,10 +24,19 @@ namespace xpyt
     wasm_interpreter::wasm_interpreter()
         : interpreter(true, true)
     {
+        m_release_gil_at_startup = false;
```
martinRenou, line 27: Do we still need this? I see you do the acquire in the `configure_impl` method

DerThorsten: yeah...this is actually needed.
All the calls in the async-event-loop from `pyjs` are done via `setTimeout`.
within `setTimeout` one is in a completely different scope as the `configure_impl` or the normal scope of the execution of cells.
Even when adding explicit code to acquire the gil in the callbacks which are passed to `setTimeout`, it still requires that change.
Since we only have a single thread anyhow there is not need for ever releasing the gil at any time.

DerThorsten:
found a tiny issue! do not merge yet!

src/xinterpreter_wasm.cpp
martinRenou: Right I guess we can keep if for now then 👍🏽

src/xinterpreter_wasm.cpp
```diff
@@ -24,10 +24,46 @@ namespace xpyt
...
     }
 
+    void wasm_interpreter::configure_impl()
+    {
+        interpreter::configure_impl();
+
+        py::gil_scoped_acquire acquire;
+        py::module sys = py::module::import("pyjs");
+
+        py::exec(R"""(
```
DerThorsten, line 41: in the next iteration  we can put this in `xeus-python-shell`

martinRenou:
Thanks!

PR #545, opened 2022-06-22
reduced inital memory and stack size

DerThorsten:
With this, more kernels can be started simultaneously.
Good chance this fixes https://github.com/jupyterlite/xeus-python-kernel/issues/37


martinRenou (approved):
Thanks!

PR #550, opened 2022-08-23
update pyjs to 0.12.0

DerThorsten:
update to new pyjs version  once https://github.com/emscripten-forge/recipes/pull/186 has landed


martinRenou:
There is another issue (probably with the way we link against libbz2 in pyjs?). I'm pinning pyjs for now in the CI and releasing. Will also pin in recipes. It's probably good to pin anyway, that way can safely make other breaking changes in pyjs without breaking xeus-python.

DerThorsten:
we might need to linnk against `libbz2` in xeus-python (or do it somehow different in pyjs)

DerThorsten:
probably it is enough to add them to the ci. we do this unholy thing in the cmakelist
```
        $ENV{PREFIX}/lib/libbz2.a
        $ENV{PREFIX}/lib/libz.a
        $ENV{PREFIX}/lib/libsqlite3.a
        $ENV{PREFIX}/lib/libffi.a
```         

so I assume adding them to the env should do the trick:
```
          micromamba create -n xeus-python-build-wasm \
            --platform=emscripten-32 \
            -c https://repo.mamba.pm/emscripten-forge \
            -c https://repo.mamba.pm/conda-forge \
            --yes \
            python ipython pybind11 jedi xtl nlohmann_json pybind11_json numpy xeus "xeus-python-shell>=0.4.1" "pyjs>=0.12.1" "wasm_requests>=0.1.5"  bzip2 sqlite zlib libffi 
             ```

martinRenou:
> probably it is enough to add them to the ci. we do this unholy thing in the cmakelist

Shouldn't we use `find_package` for that?  (I'm not a cmake expert) 

> so I assume adding them to the env should do the trick:

Mmh, so maybe those are missing dependencies for the pyjs recipe?

DerThorsten:
> Shouldn't we use find_package for that? (I'm not a cmake expert)

Yeah this was a early hack which got copy-pasted a lot (probably by me)

> Mmh, so maybe those are missing dependencies for the pyjs recipe?

actually, these are python dependencies. But we only need those when we need to build something with python and use the libpython.
When we have a pkg like numpy which also depends on python we do not want these dependencies. this is why there are not part of the run-dependencies.

The solution would be to have another pkg, smth like "libpython" which has these as dependencies.

PR #559, opened 2022-12-05
Fix rich inspect variables in function breakpoint

### LINKED ISSUE #558: Rich variable inspector fails in a function frame
The rich variable inspector in JupyterLab debugger does not work if the current frame is a function:

image

The variable is never shown, for global or local variables.

Everything works fine again if the frame is the module:

image

brichet:
Fixes #558

Uses an evaluate request with the clipboard context to get the variable representation, and stores the result in global variables using an `exec` command.
This is based on what is done in `ipykernel` to get the variable representation.

**NOTE:** There may be a simplest way (which I haven't found) to save result values in 2 global variables, instead of executing code that evaluates the result...

### DIFF
--- a/src/xdebugger.cpp
+++ b/src/xdebugger.cpp
@@ -122,21 +122,25 @@ namespace xpyt
         {
             // The code has stopped on a breakpoint, we use the setExpression request
             // to get the rich representation of the variable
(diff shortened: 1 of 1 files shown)


JohanMabille:
Thanks!

PR #562, opened 2022-12-09
Add copy_to_globals request

brichet:
This PR adds a new request to copy a variable to the global scope, and tests on the feature.

It's a draft PR until https://github.com/microsoft/debugpy/pull/1147 is released.

### DIFF
--- a/include/xeus-python/xdebugger.hpp
+++ b/include/xeus-python/xdebugger.hpp
@@ -71,7 +71,7 @@ namespace xpyt
         std::string m_debugpy_port;
         nl::json m_debugger_config;
         py::object m_pydebugger;
-        std::string m_debugpy_version;
+        bool m_version_lt_1_6_5;
     };
 
     XEUS_PYTHON_API
--- a/include/xeus-python/xdebugger.hpp
+++ b/include/xeus-python/xdebugger.hpp
@@ -71,6 +71,7 @@ namespace xpyt
         std::string m_debugpy_port;
         nl::json m_debugger_config;
         py::object m_pydebugger;
+        std::string m_debugpy_version;
     };
 
     XEUS_PYTHON_API
--- a/include/xeus-python/xdebugger.hpp
+++ b/include/xeus-python/xdebugger.hpp
@@ -71,7 +71,7 @@ namespace xpyt
         std::string m_debugpy_port;
         nl::json m_debugger_config;
         py::object m_pydebugger;
-        bool m_version_lt_1_6_5;
+        bool m_copy_to_globals_available;
     };
 
     XEUS_PYTHON_API
--- a/include/xeus-python/xdebugger.hpp
+++ b/include/xeus-python/xdebugger.hpp
@@ -54,6 +54,7 @@ namespace xpyt
         nl::json rich_inspect_variables_request(const nl::json& message);
         nl::json attach_request(const nl::json& message);
         nl::json configuration_done_request(const nl::json& message);
+        nl::json copy_to_globals_request(const nl::json& message);
 
         nl::json variables_request_impl(const nl::json& message) override;
 
--- a/src/xdebugger.cpp
+++ b/src/xdebugger.cpp
@@ -195,14 +195,14 @@ namespace xpyt
     nl::json debugger::copy_to_globals_request(const nl::json& message)
     {
         // This request cannot be processed if the version of debugpy is lower than 1.6.5.
-        if (less_than_version(m_debugpy_version, "1.6.5"))
+        if (m_version_lt_1_6_5)
         {
             nl::json reply = {
                 {"type", "response"},
                 {"request_seq", message["seq"]},
                 {"success", false},
                 {"command", message["command"]},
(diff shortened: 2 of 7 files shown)


include/xeus-python/xdebugger.hpp
```diff
@@ -71,6 +71,7 @@ namespace xpyt
         std::string m_debugpy_port;
         nl::json m_debugger_config;
         py::object m_pydebugger;
+        std::string m_debugpy_version;
```
JohanMabille, line 74: I think storing a boolean stating if the `copy_to_globals_request` is available might be more efficient.

src/xdebugger.cpp
```diff
@@ -273,6 +286,10 @@ namespace xpyt
             py::gil_scoped_acquire acquire;
             py::module xeus_python_shell = py::module::import("xeus_python_shell.debugger");
             m_pydebugger = xeus_python_shell.attr("XDebugger")();
+
+            std::string expression = "debugpy.__version__";
+            py::object version = eval(py::str(expression));
+            m_debugpy_version = version.cast<std::string>();
```
JohanMabille, line 292: A good idea here would be to replace the `.` char in `m_debug_py` with spaces, to ease future comparison:

```cpp
auto debugpy_version = version.cast<std::string>();
std::replace(debugpy_version.begin(), debugpy_version.end(), '.' , ' ');
```

You can then initialize the boolean member by calling a comparison function with debugpy_version and the minimal version of debugpy in the same format (i.e. "1 6 5");

src/xutils.cpp
```diff
@@ -149,4 +149,38 @@ namespace xpyt
...
+            result.push_back(value);
+            parser.get();
+        }
+        return result;
+    }
+
+    // Compare 2 version and return true if version1 < version2.
+    // The versions must be strings formatted as the regex: [0-9]+(\.[0-9]+)*
+    bool less_than_version(std::string version1, std::string version2)
+    {
```
JohanMabille, line 176: If we assume that `version1` and `version2` have the format "MAJOR MINOR PATCH", then the implementation of this function can be improved:

```cpp
using iterator_type = std::istream_iterator<int>;
std::istringstream iv1(version1), iv2(version2);
return std::lexicographical_compare(
    iterator_type(iv1),
    iterator_type(),
    iterator_type(iv2),
    iterator_type()
);
```

JohanMabille (changes requested):

include/xeus-python/xdebugger.hpp
brichet: OK, I thought we could keep it generic for future changes.

src/xdebugger.cpp
brichet: I don't see the difference if we replace '.' with ' '.
Can it simplify the comparison ?

JohanMabille: Yes because we can then use the istream_iterator as a tokenizer (it skips the spaces in a string by default, see the implementation in the comment below).

include/xeus-python/xdebugger.hpp
```diff
@@ -71,7 +71,7 @@ namespace xpyt
         std::string m_debugpy_port;
         nl::json m_debugger_config;
         py::object m_pydebugger;
-        std::string m_debugpy_version;
+        bool m_version_lt_1_6_5;
```
JohanMabille, line 74: I think that `m_copy_to_globals_available` would be more explicit, the fact that it is available onl after version 1.6.5 is an implementation detail.

src/xutils.cpp
brichet: Thanks @JohanMabille.
Not sure we can assume it as users can still have alpha/beta versions I thinks.

JohanMabille: In that case we need to split the string to remove the alpha / beta part.

include/xeus-python/xdebugger.hpp
brichet: right, thanks.

src/xutils.cpp
brichet: OK, let's do it this way

src/xdebugger.cpp
```diff
@@ -287,9 +287,20 @@ namespace xpyt
             py::module xeus_python_shell = py::module::import("xeus_python_shell.debugger");
             m_pydebugger = xeus_python_shell.attr("XDebugger")();
 
+            // Get debugpy version
             std::string expression = "debugpy.__version__";
             std::string version = (eval(py::str(expression))).cast<std::string>();
-            m_copy_to_globals_available = less_than_version(version, "1.6.5");
+
+            // Format the version to match [0-9]+(\s[0-9]+)*
+            int pos = version.find_first_of("abrc");
```
JohanMabille, line 295: `find_first_pos` returns an unsigned int, so this should be `size_t pos = ...` . Then pos must be compared to `std::string::npos`: `if (pos != std::string::npos) ... `

brichet: Thanks, I first tried it but I was comparing it to `-1` and the compiler was not OK :smile:

JohanMabille (approved):

JohanMabille:
Thanks!

PR #573, opened 2023-02-06
Pyjs 0 18

DerThorsten:


environment-wasm-host.yml
```diff
@@ -14,5 +14,9 @@ dependencies:
...
   - xeus 
   - xeus-python-shell>=0.5
-  - pyjs>=0.7.0,<0.8
-  - wasm_requests>=0.1.5
+  - pyjs>=0.18.2
+  - requests-wasm-polyfill>=0.1.5
+  - bzip2 
+  - sqlite 
+  - zlib 
+  - libffi
```
JohanMabille, line 22: Why do we need these additional dependencies?

DerThorsten:
we need this whenever we *link* against lib-python.
But we cannot / should-not make this a run dependency of python. Otherwise all side-packages will have these libraries as dependencies

DerThorsten:
@JohanMabille we can use `libpython` instead of `python` https://beta.mamba.pm/channels/emscripten-forge/packages/libpython which has these additional pkgs as dependencies

JohanMabille:
Ok, thanks for the explanation. And +1 for the meta package!

PR #574, opened 2023-02-06
bumped patch

DerThorsten:


JohanMabille:
We push release commit directly to master, no need for a PR.

PR #588, opened 2023-04-21
require `pyjs >= 1.0.0`

DerThorsten:
I also removed empack from the CI which was used to pack the env in a single js/data blob. 
Since the single js/data blobs are outdated anyhow I removed that part

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -193,15 +193,14 @@ jobs:
         run: |
           micromamba activate xeus-python-wasm-build
 
-          ./$HOME/emsdk activate ${{matrix.emsdk_ver}}
-          source $CONDA_EMSDK_DIR/emsdk_env.sh
+          $HOME/emsdk activate ${{matrix.emsdk_ver}}
+          source $HOME/emsdk/emsdk_env.sh
 
           micromamba create -f environment-wasm-host.yml --platform=emscripten-32
 
           mkdir build
           pushd build
 
-          export EMPACK_PREFIX=$MAMBA_ROOT_PREFIX/envs/xeus-python-wasm-build
           export PREFIX=$MAMBA_ROOT_PREFIX/envs/xeus-python-wasm-host
           export CMAKE_PREFIX_PATH=$PREFIX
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yml
```diff
@@ -206,20 +211,13 @@ jobs:
...
 
-          make -j5
-
-          empack pack env --env-prefix $PREFIX --outname python_data --config $EMPACK_PREFIX/share/empack/empack_config.yaml
+          make -j5 
 
-          popd
-
-          # Patch output
-          python wasm_patches/patch_it.py
```
martinRenou, line 216: Is the patch not needed anymore?

DerThorsten: To be honest, I am not 100% sure. At least for the pure pyjs version I use in the pyjs-code-runner does require this patch. 
We can also leave it in.
What is the intent of uploading the files? are the currently  consumed by any of our tools?

DerThorsten: Ok,then we leave it in, but I make an issue to check if we actually still need it

DerThorsten: > That was to be able to test them manually in jupyterlite-xeus-python prior to releasing, which was actually not possible without the relocate feature. But it may be possible now?

So the `libpython` we get in the ci has the prefix `$MAMBA_ROOT_PREFIX/envs/xeus-python-wasm-host` ie smth like `/home/runner/micromamba/envs/xeus-python-wasm-host`. So the generated xeus wasm will have this prefix baked in.

So if you want to use these artifacts you need to create the env with `micromamba` and relocated it to
`/home/runner/micromamba/envs/xeus-python-wasm-host`
This might work.

DerThorsten: https://github.com/jupyter-xeus/xeus-python/issues/589

@martinRenou  btw I supsect if you just pass the `relocate-prefix` argument  `empack pack env` it will just work. even if the env was not relocated!  This can work because I explicitly set `PYTHONHOME` to the relocated prefix in pyjs
https://github.com/emscripten-forge/pyjs/blob/main/include/pyjs/pre_js/init.js#L25-L26

martinRenou (approved):
Thanks!

Issue #596, opened 2023-11-07
merging `xeus-python-kernel` into this repo


DerThorsten:
Atm we have two repositories for the `xeus-python` project:

**1** https://github.com/jupyter-xeus/xeus-python
**2** https://github.com/jupyterlite/xeus-python-kernel
 
That leads to the following problems:
To test a single change in the `xeus-python` project, we need to jump through a lot of hoops.

 * create a new tag / version in the `xeus-python` project.
 * release the new tag / version on `emscripten-forge` (or do a local build of an `emscripten-forge` package)
 * update the `xeus-python` version in the `xeus-python-kernel` project to use the new version.
 * build the `xeus-python-kernel` project.
 * see if the changes work

The testing/ci  on the `xeus-python` project is non-existent (wrt wasm). While on the `xeus-python-kernel` project, we
have the mechanism to run ui tests for the `xeus-python-kernel` on each commit.

Therefore it would be beneficial to merge the `xeus-python` and `xeus-python-kernel` projects into a single project.
The benefits are:
 
* we can test the `xeus-python` project on each commit 
* we can iterate faster on the `xeus-python` project.
* more robust

The downsides:
  
* The repo will be bigger.

I believe the benefits outweigh the downsides massively.

Furthermore, In the long-run I think this merging of `xeus-<lang>` and `xeus-<lang>-kernel` projects should be done for all
`xeus-<lang>` projects. This also allows to have a single cookiecutter / template repo for new `xeus-<lang>` projects.
ATM I beieve no "outsider" can create a new `xeus-<lang>` lite kernel, because the process is too complicated.

Therefore, I suggest that we create a folder `lite` in the `xeus-python` project, where we put the `xeus-python-kernel` project.
We need to adapt the npm build system to build the `xeus-python` lite from source.
We maybe want to keep the ability to build the `xeus-python` lite kernel from a emscripten-forge pre-built `xeus-python` package.

Any suggestions on that? @martinRenou @JohanMabille @SylvainCorlay @jtpio

martinRenou:
Thank for starting this discussion @DerThorsten. I would be in favor of this 👍🏽 💯 Especially that it would make the jupyterlite-xeus-python kernel more robust by catching issues earlier.

JohanMabille:
I'm in favor of this merge too, with the following remarks:

[...]

PR #599, opened 2023-11-07
Merge xeus python lite2

DerThorsten:


martinRenou (approved):
🚀 Thanks!

jtpio:
Is there an issue to track the remaining items to be done after the incorporation in the main repo?

lite/.github/workflows/build.yml
```diff
@@ -0,0 +1,212 @@
+name: Build
```
jtpio, line 1: These workflows should be hoisted to the top-level so they can actually run.

lite/docs/build-environment.yml
```diff
@@ -0,0 +1,21 @@
+name: xeus-python-kernel-docs
```
jtpio, line 1: Same as above, this should be merged with the existing docs.

lite/.github/workflows/build.yml
martinRenou: The work is being done in separate PRs against the [merge-xeus-python-kernel](https://github.com/jupyter-xeus/xeus-python/tree/merge-xeus-python-kernel) branch

jtpio:
> Is there an issue to track the remaining items to be done after the incorporation in the main repo?

It's in https://github.com/jupyter-xeus/xeus-python/issues/596

PR #600, opened 2023-11-08
added emsdk and build script

DerThorsten:


lite/jupyterlite_xeus_python/build.py
```diff
@@ -39,6 +39,17 @@
...
+# check if we are in a local build via
+# export XEUS_PYTHON_LOCAL_BUILD=1
+# export XEUS_PYTHON_BUID_DIR=$XEUS_PYTHON_BUILD_DIR
+XEUS_PYTHON_LOCAL_BUILD=False
+if os.environ.get("XEUS_PYTHON_LOCAL_BUILD") == "1":
+    XEUS_PYTHON_LOCAL_BUILD=True
+    XEUS_PYTHON_BUILD_DIR=os.environ.get("XEUS_PYTHON_BUILD_DIR")
+    if XEUS_PYTHON_BUILD_DIR is None:
+        raise RuntimeError("XEUS_PYTHON_BUILD_DIR is not set")
+    XEUS_PYTHON_BUILD_DIR=Path(XEUS_PYTHON_BUILD_DIR)
```
martinRenou, line 51: Could we make the build directory as part of the Python package data (local dir installed alongside the Python package under site-packages)?

lite/build.sh
```diff
@@ -0,0 +1,110 @@
...
+source $EMSDK_INSTALL_DIR/emsdk_env.sh --deactivate
+
+#############################
+# Build Python
+############################
+pushd $LITE_SRC_DIR
+export XEUS_PYTHON_LOCAL_BUILD=1
+export XEUS_PYTHON_BUILD_DIR=$XEUS_PYTHON_BUILD_DIR
+echo "building python package"
+python -m pip install -e ".[dev]"
```
martinRenou, line 101: `pip install .` already calls `build.py`, how about putting the content of you `build.sh` in `build.py`? That way building a wheel with `python -m build` would do everything required to build xeus-python WASM as well.

jtpio:
@DerThorsten the "Enforce PR label" is for making sure each PR has a GitHub label, so the releaser changelog is automatically nicely formatted. 

For example for this PR you could add "enhancement" (I don't have the rights to do it on this repo)

jtpio:
"Check Release" is to make sure the repo is always in a state where it can be released. Maybe it can be ignored for now as it's not cleared how versioning and releasing should be done now that this repo hosts multiple packages (as discussed a bit in https://github.com/jupyter-xeus/xeus-python/issues/596)

DerThorsten:
I would say we stop this effort and focus only on the plan  described in https://github.com/jupyter-xeus/xeus-lite/issues/7

PR #605, opened 2023-12-07
make wasm build more consistent

DerThorsten:
* Using the same name for the binary in the wasm case allows us to keep the `kernel.json` as it used to be, ie no more `wasm_file` entry in the metadata. When we need to figure out where the wasm/js is, we can just take `argv[0]` from the kernel.json.
* Install wasm binaries to `$PREFIX/bin` instead `$PREFIX/share/jupyter/kernels/xpython`


martinRenou (approved):
LGTM! Thanks

PR #607, opened 2024-01-05
Update docs showcasing jupyterlite xeus-python

martinRenou:


martinRenou:
@DerThorsten this is working! Using the jupyterlite-xeus alpha release on Pypi. You can try it here: https://xeus-python--607.org.readthedocs.build/en/607/

A couple of comments/questions though:
- The kernel is named "Python 3.12 (XPython)" but it's actually Python 3.11 that is installed, we should maybe find the reason for that
- This is not yet using a local build of xeus-python, we should try to build it on readthedocs (if there is enough memory we can use there)

DerThorsten:
> * The kernel is named "Python 3.12 (XPython)" but it's actually Python 3.11 that is installed, we should maybe find the reason for that

https://github.com/jupyter-xeus/xeus-python/blob/main/share/jupyter/kernels/xpython/kernel.json.in#L2C28-L2C48 I assume the python in the build environment was probably python 3.12 ...at least that would be my first guess

martinRenou:
Doesn't it mean that `PYTHON_VERSION_MINOR` is not set properly during the wasm build on emscripten-forge?

DerThorsten:
I would rather blame some FindPython.cmake script we use when building xeus-python.
Probably somewhere in the pybind11 logic build and host python are confused

PR #609, opened 2024-01-11
fix wrong python version in kernelspec in wasm case

DerThorsten:
* the python version from build python instead of host/wasm python wasm shown in kernelspec


martinRenou (approved):
Thanks!

martinRenou:
I will publish a release of xeus-python

PR #610, opened 2024-01-17
catch cpp exceptions

DerThorsten:
I noticed that C++ exceptions thrown from a Python package built with pybind11 are not handled properly.
In the jupyterlite case, they only appear in the browser console but not in the notebook as expected.

This simple try/catch block fixes the issue. 
I also removed the unused ` py::object ipython_res`


martinRenou (approved):
Thanks!

PR #615, opened 2024-03-04
Removing GIL

IsabelParedes:


martinRenou:
@IsabelParedes I've not followed the discussions around the removal of the GIL. Can you provide a bit more context as to why try to remove the GIL scope acquirement? Also why are we not going forward with this?

PR #616, opened 2024-03-08
update jupyter lite link

shimwell:
current link goes to a deprecate repo so I thought might be worth updating the link


martinRenou (approved):
Thanks!

PR #617, opened 2024-03-13
use pyjs 2.0.0 

DerThorsten:
* the only breaking chage is, the embedded module needs to be named pyjs_core instead of pyjs


martinRenou (approved):
Thanks!

PR #618, opened 2024-03-13
Minor fixes

IsabelParedes:
Fixes these warnings:

```sh
warning: conversion from 'size_t' {aka 'long unsigned int'} to 'int' may change value [-Wconversion]
  120 |         backtrace_symbols_fd(array, size, STDERR_FILENO);
```

```sh
warning: 'void zmq::detail::socket_base::setsockopt(int, const void*, size_t)' is deprecated: from 4.7.0, use `set` taking option from zmq::sockopt [-Wdeprecated-declarations]
```

### DIFF
--- a/src/xutils.cpp
+++ b/src/xutils.cpp
@@ -113,7 +113,7 @@ namespace xpyt
         void* array[10];
 
         // get void*'s for all entries on the stack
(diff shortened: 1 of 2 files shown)


martinRenou (approved):
Thanks!

Issue #568, opened 2023-01-10
Py_SetPythonHome is deprecated


SylvainCorlay:
We are seeing the following error message in CI:

``` 
/src/xpaths.cpp:75:60: warning: 'void Py_SetPythonHome(const wchar_t*)' is deprecated [-Wdeprecated-declarations]
   75 |         Py_SetPythonHome(const_cast<wchar_t*>(wstr.c_str()));
```

anutosh491:
Can be closed as https://github.com/jupyter-xeus/xeus-python/pull/629 is now merged

PR #629, opened 2024-06-11
Update deprecated functions

IsabelParedes:
- `find_package(PythonInterp)` is deprecated since cmake v3.12 (See [FindPythonInterp](https://cmake.org/cmake/help/latest/module/FindPythonInterp.html)).

- The following have been deprecated since Python 3.11 and are replaced with a `PyConfig` structure introduced in Python 3.8
  - `Py_SetProgramName`
  - `Py_SetPythonHome`
  - `PySys_SetArgvEx`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,12 +8,12 @@
 # The full license is in the file LICENSE, distributed with this software. #
 ############################################################################
 
-cmake_minimum_required(VERSION 3.10)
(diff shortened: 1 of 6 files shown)


anutosh491:
I think this should fix https://github.com/jupyter-xeus/xeus-python/issues/568

IsabelParedes:
`macos-11` deprecated sooner than announced.

JohanMabille (approved):

PR #630, opened 2024-06-17
Update CI

IsabelParedes:
- `macos-11` no longer runs. It will be officially deprecated 28.06.2024
- `SimenB/github-actions-cpu-cores` does not work


martinRenou (approved):
Thanks!

PR #631, opened 2024-06-19
Removing dependency on cppzmq

anutosh491:
I guess the dependency on cppzmq was dropped after migration to xeus 5.


anutosh491:
Not sure why one of the `test-win` fails. Maybe someone can re-run the failed job

PR #640, opened 2024-09-25
Passing the link of empack_env_meta.json file

AnastasiaSliusar:
This PR is the part of the solution of https://github.com/QuantStack/qs.ai/issues/28
It replaces the hardcoded meantion  of "empack_env_meta.json" file by a variable where this file is located. 

It should be used with https://github.com/jupyterlite/xeus/pull/120 and https://github.com/emscripten-forge/pyjs/pull/69 together

### DIFF
--- a/wasm_patches/post.js
+++ b/wasm_patches/post.js
@@ -22,11 +22,10 @@ Module['async_init'] = async function(
      packages_json_url = empack_env_meta_link;
     }
      
-
     Module['bootstrap_from_empack_packed_environment']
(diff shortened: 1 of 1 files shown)


trungleduc:
I don't think this PR is needed unless we want to relax the hard-coded `empack_env_meta.json` file name. Using `kernel_root_url` is already enough to customize the file location. What is your opinion @martinRenou @DerThorsten?

martinRenou:
I agree, I'd be in favor of reducing the diff as much as possible, and this PR is indeed not mandatory for what we want to achieve

PR #641, opened 2024-10-18
Add `zstd` for wasm build

trungleduc:
Needed for https://github.com/emscripten-forge/pyjs/pull/76

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -382,6 +382,7 @@ if (EMSCRIPTEN)
         $ENV{PREFIX}/lib/libz.a
         $ENV{PREFIX}/lib/libsqlite3.a
         $ENV{PREFIX}/lib/libffi.a
+        $ENV{PREFIX}/lib/libzstd.a
     )
 
     add_executable(xpython ${XPYTHON_WASM_SRC})


trungleduc:
@martinRenou @JohanMabille do you know if the failed tests are related to my changes?

anutosh491:
It seems the error comes out of this test
https://github.com/jupyter-xeus/xeus-python/blob/87251496038177d05770e3623494547517ec969d/test/test_debugger.cpp#L1246

If you need this to go in, maybe you could comment this and add a TODO ?

anutosh491:
It seems some specific operations might be going wrong (first `external_next_continue` fails and now `next_continue`)
We had all tests working as expected during the last release. Haven't looked into it but not sure why some tests fail now. 

@trungleduc could you check and add todo's for the failing ones. I shall look into them once that is done !

anutosh491:
I guess we can do what you did, what I meant was to only comment out the tests that fail (I guess a couple would fail) and add todo's for what fails

Otherwise a better way to do this might be to comment this out and add a todo
https://github.com/jupyter-xeus/xeus-python/blob/87251496038177d05770e3623494547517ec969d/.github/workflows/main.yml#L71-L74

And these out 
https://github.com/jupyter-xeus/xeus-python/blob/87251496038177d05770e3623494547517ec969d/test/CMakeLists.txt#L53-L55

So that we don't end up with this
```
Run ./test_xeus_python
[doctest] doctest version is "2.4.11"
[doctest] run with "--help" for options
===============================================================================
[doctest] test cases: 0 | 0 passed | 0 failed | 0 skipped
[doctest] assertions: 0 | 0 passed | 0 failed |
[doctest] Status: SUCCESS!
```

trungleduc:
@anutosh491 is it ok now?

JohanMabille (approved):
The failures of the debugger tests are not related to your changes. I think we can merge this one and investigate the failing tests in a dedicated PR.

PR #623, opened 2024-05-21
Upgraded to jupyter_kernel_test 0.7.0

JohanMabille:
This is a PR for testing different upgrade, it should not be merged.

PR #648, opened 2025-01-12
Update jupyter_kernel_test to 0.7

anutosh491:


anutosh491:
Hey @JohanMabille 

I see #623 was made in the same regard (not sure if it was meant for an udpate or only limited to testing) but I think we can move to a later version if we want to. Let me know if we shouldn't be making this change.

JohanMabille:
The tests on #623 were failing for a unknwown reason, and I did not have the time to investigate. Since things seem to work now, let's metge this and close #623!

PR #656, opened 2025-05-06
Link to the Jupyter Zulip

jtpio:
Since Xeus is a Jupyter subproject, and Gitter is not being actively used anymore, we can update the badge and direct people to the Jupyter Zulip instance.


martinRenou (approved):
Thanks!

PR #657, opened 2025-07-01
Switch to a shared xeus for the emscripten build

anutosh491:
Replicating the PR on xeus-cpp so that xeus-python can use the latest xeus & xeus-lite.

https://github.com/compiler-research/xeus-cpp/pull/334

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,16 +44,23 @@ if(EMSCRIPTEN)
     STRING(REGEX MATCH "python[0-9]+[.][0-9]+" PYTHON_VERSION_STRING ${WASM_PYTHON_LIBRARY})
     STRING(REGEX MATCH "[0-9]+" PYTHON_VERSION_MAJOR ${PYTHON_VERSION_STRING})
     STRING(REGEX MATCH "[0-9]+$" PYTHON_VERSION_MINOR ${PYTHON_VERSION_STRING})
+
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -44,16 +44,23 @@ if(EMSCRIPTEN)
     STRING(REGEX MATCH "python[0-9]+[.][0-9]+" PYTHON_VERSION_STRING ${WASM_PYTHON_LIBRARY})
     STRING(REGEX MATCH "[0-9]+" PYTHON_VERSION_MAJOR ${PYTHON_VERSION_STRING})
     STRING(REGEX MATCH "[0-9]+$" PYTHON_VERSION_MINOR ${PYTHON_VERSION_STRING})
+
+    configure_file (
+        "${CMAKE_CURRENT_SOURCE_DIR}/share/jupyter/kernels/xpython/wasm_kernel.json.in"
+        "${CMAKE_CURRENT_SOURCE_DIR}/share/jupyter/kernels/xpython/kernel.json"
+    )
```
anutosh491, line 51: Fetching xeus-python from emscripten-forge, I see that we're only interested in the xpython build and not xpython-raw so we can skip configuring it for the wasm use case.

anutosh491:
cc @martinRenou

martinRenou (approved):
Thanks!

PR #673, opened 2025-10-15
emscripten-4x + 3x support

DerThorsten:


DerThorsten:
> We still keep the post.js in the kernels correct ?

yes, no need to get rid of them **right now**, but in the longrun we may phase them out

PR #674, opened 2025-10-17
Add Modules Request to the debugger protocol

arjxn-py:
Adds Modules request to the debugger protocol

### DIFF
--- a/src/xdebugger.cpp
+++ b/src/xdebugger.cpp
@@ -374,6 +374,7 @@ namespace xpyt
 
     nl::json debugger::modules(const nl::json& message)
     {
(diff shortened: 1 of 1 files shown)


JohanMabille (commented):
Thanks for tackling this! The implementation is neat (although I wonder if we should have it in Python in [xeus-python-shel](https://github.com/jupyter-xeus/xeus-python-shell), and have this method just forward the call).

You need to register the function as the message handler [here](https://github.com/jupyter-xeus/xeus-python/blob/main/src/xdebugger.cpp#L68), and declrare the method in the debugger header.

Issue #522, opened 2022-01-13
Use xeus's routines for creating the replies


martinRenou:
As suggested by @DerThorsten we should use the `xeus`'s routines for creating the replies, this would prevent some issues we faced with missing entries in the replies.

PR #675, opened 2025-10-21
Use xeus's routines for creating the replies

arjxn-py:
Fix https://github.com/jupyter-xeus/xeus-python/issues/522

### DIFF
--- a/src/xinterpreter.cpp
+++ b/src/xinterpreter.cpp
@@ -121,7 +121,6 @@ namespace xpyt
                                            nl::json user_expressions)
     {
         py::gil_scoped_acquire acquire;
-        nl::json kernel_res;
 
         // Reset traceback
         m_ipython_shell.attr("last_error") = py::none();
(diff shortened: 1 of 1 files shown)


arjxn-py:
cc: @JohanMabille @martinRenou 

Is this how we want to use the xeus's routines?

src/xinterpreter.cpp
```diff
@@ -270,19 +270,6 @@ namespace xpyt
...
-
-        /* The jupyter-console banner for xeus-python is the following:
-          __  _____ _   _ ___
-          \ \/ / _ \ | | / __|
-           >  <  __/ |_| \__ \
-          /_/\_\___|\__,_|___/
-
-          xeus-python: a Jupyter lernel for Python
-        */
-
```
martinRenou: Can we keep this comment?

martinRenou (approved):
Thanks!

